PR #11, opened 2019-10-25
RSE-306: Add Award Basic Details API

tunbola:
## Overview
This PR adds the Entity and API for saving an Award detail. 
An Award detail has the following fields:

- Award type
- Start date
- End date
- Award Managers

The Profile ID field added will be used for storing custom fields in a subsequent PR.

## Before
The Award Detail Entity does not exist.

## After
An Award detail is simply an extension of the information stored for a case Type. More like extra fields. So we have the case_type_id column on the `civicrm_civiawards_award_detail` table.

The AwardDetail Entity was created

Table Info: civicrm_civiawards_award_detail
  `id` int
  `case_type_id` int
  `award_type` : One of the values of the award_type option group',
  `start_date` 'Award Start Date',
  `end_date` 'Award End Date',
  `profile_id` int: Will be used to store Award custom fields,

AwardManager Entity was created

Table info:  `civicrm_civiawards_award_manager`
  `id` int
  `contact_id` int,
  `case_type_id` int

## Technical Details
Two entities AwardDetail and AwardManager was created to store the Award basic details. The AwardDetail.create API is exposed to the frontend to store all these fields as these details are created together on the Award details form. Behind the scene the AwardDetail API stores and retrieves the Award manager information from the Award Manager entity. 
The Award managers are stored in a separate table rather than a list or array on the Award Details table because there will be need to fetch a list of all distinct Award managers later and also this table structure will be useful incase we need to introduce reports later on.

**Creating an Award Detail:**

```php
$result = civicrm_api3('AwardDetail', 'create', [
  'award_manager' => [202, 203],
  'case_type_id' => 3,
  'start_date' => "2019-06-01",
  'end_date' => "2019-06-02",
  'award_type' => "prizes",
]);
```

**Deleting an Award Detail**:
Deleting an award detail will also delete then corresponding award manager details for the case type behind the scene.

```php
$result = civicrm_api3('AwardDetail', 'delete', [
  'id' => 4],
]);
```

## Comments
The Award Type option group will be added in a later PR.


ang/civiawards/awards/AwardsNew.html
```diff
@@ -32,8 +32,8 @@ <h1 crm-page-title>{{title}}</h1>
           </label>
           <div class="col-sm-5">
             <input class="form-control"
-              ng-model="selectedActivityType"
-              crm-ui-select="{
+                   ng-model="selectedActivityType"
```
MiyaNoctem: This file seems to only have whit-space changes, is this intentional?

tunbola: I merged this in from the `RSE-305-create-an-award` branch. It didn't have a conflict so that must be how it is in original branch.

PR #12, opened 2019-10-25
RSE-492: Create award type option group.

tunbola:
## Overview
This PR adds the Award Type option group with default option values.

## Before
This option group does not exist.

## After
The option group is created

Option Groups  case-prospect 2019-10-25 14-38-29


deb1990:
@tunbola Code looks good, but will this run when creating a new site?

tunbola:
@deb1990 , Yes it will run on new site install since the setup is added to the install function in the Awards Upgrader file. You can verify by deploying on compudeploy if you wish.

PR #13, opened 2019-10-28
RSE-347: Create/Edit Awards Functionality

deb1990:
## Overview
As part of this PR, the Create/Edit Awards Functionality has been created.
Related PR:
https://github.com/compucorp/uk.co.compucorp.civicrm.prospect/pull/43
https://github.com/compucorp/uk.co.compucorp.civicase/pull/284

## Screenshot
local site_civicrm_a_

## Technical Details
1. `awardTypes` option group is exposed to Frontend in `civiawards.ang.php` using `CRM.civiawards.awardTypes` global variable. `award-type.service.js` is created to wrap this inside `AwardType` service.

2. In `app.routes.js`, route for Editing awards is added
```javascript
$routeProvider.when('/awards/:awardId', {
  template: function (params) {
    return '<civiaward award-id="' + params.awardId + '"></civiaward>';
  }
});
```

3. `civiaward` directive is created, which wraps all the functionality for adding/editing the award.

4. Unit tests support and unit tests are added.


ang/civiawards/directives/award.directive.js
```diff
@@ -0,0 +1,231 @@
...
+      params.title = $scope.basicDetails.title;
+      params.description = $scope.basicDetails.description;
+      params.is_active = $scope.basicDetails.isEnabled;
+      params.case_type_category = awardsCaseTypeCategoryValue;
+      params.name = $scope.basicDetails.name;
+      params.definition = {};
+      params.definition.statuses = selectedAwardStages.length === _.size($scope.awardStages) ? [] : selectedAwardStages;
+
+      return crmApi('CaseType', 'create', params);
+    }
```
reneolivo:
```js
function saveCaseTypeBasicDetails() {
  var selectedAwardStageNames = _.chain($scope.basicDetails.selectedAwardStages)
    .filter(function (isSelected) { return isSelected; })
    .map(function (isSelected, key) { return $scope.awardStages[key].name; })
    .value();
  var areAllAwardsSelected = selectedAwardStages.length === _.size($scope.awardStages)
  var params = {
    sequential: true,
    title: $scope.basicDetails.title,
    description: $scope.basicDetails.description,
    is_active: $scope.basicDetails.isEnabled,
    case_type_category: CaseTypeCategory.findValueByName('awards'), // I'd consider adding this to the service
    name: $scope.basicDetails.name,
    definition: {
      statuses: areAllAwardsSelected ? [] : selectedAwardStages
    }
  };

  if ($scope.awardId) {
    params.id = $scope.awardId;
  }

  return crmApi('CaseType', 'create', params);
}
```

ang/test/civiawards/directives/award.directive.spec.js
```diff
@@ -0,0 +1,255 @@
...
+      $scope.basic_details_form = { awardName: { $pristine: true } };
+    }));
+
+    describe('basic tests', () => {
+      describe('when creating new award', function () {
+        beforeEach(() => {
+          createController({ ifNewAward: true });
+        });
+
+        it('name field is disabled', () => {
```
reneolivo: `it('disables the name field', () => {`

ang/test/civiawards/directives/award.directive.spec.js
```diff
@@ -0,0 +1,255 @@
...
+          expect($scope.isNameDisabled).toBe(true);
+        });
+
+        it('submit button is disabled', () => {
```
reneolivo: `it submits`

ang/test/civiawards/directives/award.directive.spec.js
```diff
@@ -0,0 +1,255 @@
...
+        describe('all basic fields', () => {
+          var expectedSelectedAwardStages = {};
+
+          beforeEach(function () {
+            _.each(CaseStatus.getAll(), (awardStage) => {
+              expectedSelectedAwardStages[awardStage.value] = true;
+            });
+          });
+
+          it('all basic details fields are empty', () => {
```
reneolivo: `it defines all basic details fields as empty`

ang/test/civiawards/directives/award.directive.spec.js
```diff
@@ -0,0 +1,255 @@
...
+
+        describe('name field', () => {
+          beforeEach(() => {
+            $scope.basicDetails.title = 'Some custom text';
+            $scope.$digest();
+          });
+
+          it('updating the title changes the name field too', () => {
+            expect($scope.basicDetails.name).toBe('some_custom_text');
+          });
```
reneolivo:
```js
describe('when the title changes', () => {
  beforeEach(() => {
    $scope.basicDetails.title = 'Some custom text';
    $scope.$digest();
  });

  it('updating the title changes the name field too', () => {
    expect($scope.basicDetails.name).toBe('some_custom_text');
  });
});

// ....

describe('when the name is manually changed', () => {
  // ...
});
```

Here you show that there are two things that happen. One when the title changes, the other when the name changes.

ang/test/civiawards/directives/award.directive.spec.js
```diff
@@ -0,0 +1,255 @@
...
+
+          it('updating the title does not change the name field', () => {
+            expect($scope.basicDetails.name).toBe('new_award');
+          });
+        });
+      });
+    });
+
+    describe('when saving award', () => {
```
reneolivo: `when saving the award`

ang/test/civiawards/directives/award.directive.spec.js
```diff
@@ -0,0 +1,255 @@
...
+          $scope.basicDetails.endDate = AwardAdditionalDetailsData.end_date;
+          $scope.basicDetails.awardType = AwardAdditionalDetailsData.award_type;
+
+          crmApi.and.returnValue($q.resolve({
+            values: [AwardAdditionalDetailsData]
+          }));
+          $scope.saveAward();
+        });
+
+        it('disables the save button initially', () => {
```
reneolivo: `it momentarily disables the save button` (I'd even remove momentarily and/or initially)

ang/test/civiawards/directives/award.directive.spec.js
```diff
@@ -0,0 +1,255 @@
...
+          expect($scope.submitInProgress).toBe(true);
+        });
+
+        describe('saves the details', () => {
```
reneolivo: `describe('saving the details', () => {`

ang/test/civiawards/directives/award.directive.spec.js
```diff
@@ -0,0 +1,255 @@
...
+              award_type: AwardAdditionalDetailsData.award_type,
+              award_manager: ['2', '1']
+            });
+          });
+
+          it('redirects to dashboard page', () => {
+            expect($window.location.href).toBe('/civicrm/case/a/?case_type_category=awards#/case?case_type_category=awards');
+          });
+
+          it('enables the save button finally', () => {
```
reneolivo: `it finally enables the save button` (I'd remove the word `finally`)

ang/test/civiawards/directives/award.directive.spec.js
```diff
@@ -0,0 +1,255 @@
...
+            expect($scope.submitInProgress).toBe(false);
+          });
+        });
+      });
+    });
+
+    describe('create new award stage', () => {
```
reneolivo: `describe('creating a new award stage', () => {`

ang/test/mocks/data/award-additional-details.data.js
```diff
@@ -0,0 +1,12 @@
+(function () {
+  var module = angular.module('civiawards.data');
+
+  module.constant('AwardAdditionalDetailsData', {
```
reneolivo: can we add to all of these a `MockData` suffix so we avoid collision with real services?

ang/civiawards/directives/award.directive.html
```diff
@@ -1,27 +1,46 @@
 <div id="bootstrap-theme">
-  <h1 crm-page-title>{{title}}</h1>
+  <h1 crm-page-title>{{basicDetails.title ? basicDetails.title : 'New Award'}}</h1>
```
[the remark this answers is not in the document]
deb1990: Nooo, this logic was completely wrong. Fixed it. thanks for catching.

ang/civiawards/directives/award.directive.js
```diff
@@ -0,0 +1,231 @@
...
+      $scope.basicDetails.title = caseType.title;
+      $scope.basicDetails.name = caseType.name;
+      $scope.basicDetails.description = caseType.description;
+      $scope.basicDetails.isEnabled = caseType.is_active === '1';
+
+      _.each(caseType.definition.statuses, function (stageName) {
+        var awardStage = _.find($scope.awardStages, function (stage) {
+          return stage.name === stageName;
+        });
+        $scope.basicDetails.selectedAwardStages[awardStage.value] = true;
```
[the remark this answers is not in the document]
deb1990: > what happens if selectedAwardStages already had values selected. Is it ok to not reset this var?

not sure, what you mean. This is happening on init, so how come it will already have some values.

Also refactored the function like this

```javascript
setAwardStages(caseType.definition.statuses);

/**
 * get Selected Award stages
 *
 * @param {Array} awardStages award stages fetched from api
 */
function setAwardStages (awardStages) {
  _.each(awardStages, function (stageName) {
    var awardStage = _.find($scope.awardStages, function (stage) {
      return stage.name === stageName;
    });
    $scope.basicDetails.selectedAwardStages[awardStage.value] = true;
  });
}
```

because if I set the value directly on `$scope.basicDetails.selectedAwardStages` instead of  returning another object, its more efficient.

ang/civiawards/directives/award.directive.js
```diff
@@ -0,0 +1,231 @@
...
+     *
+     * @param {object} additionalDetails additional details
+     */
+    function setAdditionalInformation (additionalDetails) {
+      $scope.awardDetailsID = additionalDetails.id;
+      $scope.basicDetails.startDate = additionalDetails.start_date;
+      $scope.basicDetails.endDate = additionalDetails.end_date;
+      $scope.basicDetails.awardType = additionalDetails.award_type;
+      $scope.basicDetails.awardManagers = additionalDetails.award_manager.toString();
+    }
```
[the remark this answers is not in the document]
deb1990: I fixed it, for all the additional details(which are not part of CaseType api), I have created `additionalDetails` scope object.

ang/civiawards/directives/award.directive.js
```diff
@@ -0,0 +1,231 @@
...
+     * Set Additional Details of award
+     *
+     * @param {object} additionalDetails additional details
+     */
+    function setAdditionalInformation (additionalDetails) {
+      $scope.awardDetailsID = additionalDetails.id;
+      $scope.basicDetails.startDate = additionalDetails.start_date;
+      $scope.basicDetails.endDate = additionalDetails.end_date;
+      $scope.basicDetails.awardType = additionalDetails.award_type;
+      $scope.basicDetails.awardManagers = additionalDetails.award_manager.toString();
```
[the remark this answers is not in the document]
deb1990: Yes the expected output for `[2,1]` is `2,1`. I can use `.join()`.

ang/civiawards.ang.php
```diff
@@ -8,6 +8,12 @@
  * http://wiki.civicrm.org/confluence/display/CRMDOC/hook_civicrm_angularModules.
  */
 
+$options = [
+  'awardTypes' => 'civiawards_award_type',
+];
+
+set_option_values_to_js_vars($options);
```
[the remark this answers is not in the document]
tunbola: Not sure if Civicrm will run the function in the parent extension first. Civi might and it might not :(. To be on the safe side, I suggest we have a `set_option_values_to_js_vars` function on this file or better still if this will end up being used by Civicase, CiviAwards, Prospects and any other Civicase extension, I suggest we extract the logic to a class in Civicase and use it anywhere we need it.
@deb1990 , @reneolivo

ang/civiawards/directives/award.directive.js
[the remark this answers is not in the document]
reneolivo: @deb1990 the problem with these type of functions is that they are modifying a variable and it's not clear that this is being done.

Performance-wise I don't think an `each` has an advantage over `reduce`. It does about the same process:
```js
function getSelectedAwards (awardStages) {
  return _.reduce(awardStages, function (selectedAwards, stageName) {
    var awardStage = _.find($scope.awardStages, function (stage) {
      return stage.name === stageName;
    });

    selectedAwards[awardStage.value] = true;

    return selectedAwards;
  }, {});
}
```

The only difference is that the final result is passed down. But I wouldn't be worried about performance here unless it's something considerable noticeable.

There's a bonus to getting the awards instead of setting them: the value gets reset each time. What happens if we call `setAwardStages` multiple times with different parameters? It would preserve the previous value. I know this is not part of the current workflow, but that's a flag.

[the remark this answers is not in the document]
deb1990: Okay makes sense. I have changed the code, but have not used `reduce`, used `each` instead.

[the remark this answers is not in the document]
shitijg: @jamienovick 

Resolved comments in the spec say that we do not need a page for Case type ie we can build on top of the current screen (which isn't the best UX).

Should create this as in the wireframe ie a bit like events?

PR #14, opened 2019-10-28
RSE-452: Implement CRUD functionality for Award Review fields

tunbola:
## Overview
Award review fields are basically custom fields from the Applicant review custom group. This PR adds the functionality to save, retrieve and delete these fields using the already available AwardDetail API  create in #11 .

## Before
CRUD functionality for Award Review fields does not exist.

## After
The award review fields can be passed in via the review fields parameter when creating an award detail.

```php
$result = civicrm_api3('AwardDetail', 'create', [
  'award_manager' => [202, 203],
  'case_type_id' => 3,
  'start_date' => "2019-06-01",
  'end_date' => "2019-06-02",
  'award_type' => "1",
  'review_fields' => [2,3,4]
]);
```

Deleting an AwardDetail will delete the review fields also
```php
$result = civicrm_api3('AwardDetail', 'delete', [
  'id' => 4],
]);
```

## Technical Details
- Review fields for an award are basically activity type custom fields. These custom fields are stored on a profile which is created when adding an Award detail. 

```php
    civicrm_api3('UFGroup', 'create', [
      'title' => $profileName,
      'name' => $profileName,
      'is_reserved' => 1,
    ]);

```
The advantages of storing the custom fields on a profile is that whenever the original custom fields are deleted, they will be automatically deleted on a profile. Also profile offers a bulk import functionality for these fields which is very useful.

- The API wrapper was also modified to return these custom fields rather than the profile Id when retrieving an Award detail

```php
civicrm_api3('AwardDetail', 'get', [
  'sequential' => 1,
  'return' => ["case_type_id"],
]);

{

    "is_error": 0,
    "version": 3,
    "count": 1,
    "id": 1,
    "values": [
        {
            "id": "1",
            "case_type_id": "1",
            "award_type": "1",
            "start_date": "2019-08-01",
            "end_date": "2019-08-02",
            "award_manager": [
                "199"
            ],
            "review_fields": [
                "2",
                "3"
            ]
        }
    ]
}
```

### DIFF
--- a/CRM/CiviAwards/Service/AwardProfile.php
+++ b/CRM/CiviAwards/Service/AwardProfile.php
@@ -90,11 +90,14 @@ class CRM_CiviAwards_Service_AwardProfile {
    *   Profile fields.
    */
   public function getProfileFields($profileID) {
+    if (empty($profileID)) {
+      return [];
+    }
     $result = civicrm_api3('UFField', 'get', [
       'uf_group_id' => $profileID,
     ]);
 
-    if ($result['count'] == 0 || empty($profileID)) {
+    if ($result['count'] == 0) {
       return [];
     }
 
--- a/CRM/CiviAwards/Service/AwardProfile.php
+++ b/CRM/CiviAwards/Service/AwardProfile.php
@@ -1,0 +1,109 @@
+<?php
+
+/**
+ * Class CRM_CiviAwards_Service_AwardProfile.
+ */
+class CRM_CiviAwards_Service_AwardProfile {
+
+  /**
+   * Create a Profile for a Case Type/Award.
+   *
+   * @param int $caseTypeId
+   *   Case Type Id.
+   *
+   * @return int
+   *   Profile id.
+   */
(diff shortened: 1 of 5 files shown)


CRM/CiviAwards/Service/AwardProfile.php
```diff
@@ -0,0 +1,109 @@
...
+   *
+   * @return array
+   *   Profile fields.
+   */
+  public function getProfileFields($profileID) {
+    $result = civicrm_api3('UFField', 'get', [
+      'uf_group_id' => $profileID,
+    ]);
+
+    if ($result['count'] == 0 || empty($profileID)) {
```
MiyaNoctem: Would it make sense to add the check for empty $profileID before you call API? I mean, is there a reason to call the API if `$profileID` is empty?

Eg.
```php
if (empty($profileID)) {
  return [];
}
$result = civicrm_api3('UFField', 'get', [
  'uf_group_id' => $profileID,
]);
```

PR #19, opened 2019-11-01
RSE-365: Add award case type configuration links

reneolivo:
## Overview
This PR adds the configuration button for award case types. These buttons are located in the overview section of the dashboard.

This PR depends on https://github.com/compucorp/uk.co.compucorp.civicase/pull/283

## Before
Screen Shot 2019-11-01 at 12 07 10 AM

## After
pr-after

## Technical Details
We add a new configuration function that adds the necessary buttons to all case types that belong to the awards category:

```js
module.config(function (CaseTypeButtonsProvider) {
  var awardCategory = getAwardCategory();
  var awardCaseTypes = getCaseTypesForCategory(awardCategory.value);

  addConfigurationButtonsToCaseTypes(awardCaseTypes);

  // ...

  function addConfigurationButtonsToCaseTypes (caseTypes) {
    _.forEach(caseTypes, function (caseType) {
      var caseTypeConfigUrl = url(AWARD_CONFIG_URL + caseType.id);

      CaseTypeButtonsProvider.addButtons(caseType.name, [{
        url: caseTypeConfigUrl
      }]);
    });
  }

  function getAwardCategory () {
    return _.find(CRM['civicase-base'].caseTypeCategories, function (category) {
      return category.name === AWARDS_CATEGORY_NAME;
    });
  }

  function getCaseTypesForCategory (categoryValue) {
    return _.filter(CRM['civicase-base'].caseTypes, function (caseType) {
      return caseType.case_type_category === categoryValue;
    });
  }
});
```

## Comments

* the `case_type_category` property of the mock award object was changed to `3` since this is the value it has on civicase for the award category.
* the `awards-case-types.config.js` config file extends the mock case types so it includes the award type as well.

### DIFF
--- a/ang/civiawards/shared/configs/case-type-buttons.config.js
+++ b/ang/civiawards/shared/configs/case-type-buttons.config.js
@@ -1,0 +1,29 @@
+(function (angular, _, url) {
+  var module = angular.module('civiawards');
+  var AWARDS_CATEGORY_NAME = 'awards';
+  var AWARD_CONFIG_URL = 'civicrm/a/#/awards/';
+
+  module.config(function (CaseTypeProvider, CaseTypeCategoryProvider, DashboardCaseTypeButtonsProvider) {
+    var awardCategory = CaseTypeCategoryProvider.findByName(AWARDS_CATEGORY_NAME);
+    var awardCaseTypes = CaseTypeProvider.getByCategory(awardCategory.value);
+
+    addConfigurationButtonsToCaseTypes(awardCaseTypes);
+
+    /**
+     * Adds configuration buttons to the given case types. The buttons url point
+     * towards the awards configuration form.
+     *
+     * @param {object[]} caseTypes a list of case types objects.
+     */
+    function addConfigurationButtonsToCaseTypes (caseTypes) {
+      _.forEach(caseTypes, function (caseType) {
+        var caseTypeConfigUrl = url(AWARD_CONFIG_URL + caseType.id);
+
+        DashboardCaseTypeButtonsProvider.addButtons(caseType.name, [{
+          icon: 'fa fa-cog',
+          url: caseTypeConfigUrl
+        }]);
+      });
+    }
+  });
+})(angular, CRM._, CRM.url);
--- a/ang/test/mocks/data/award.data.js
+++ b/ang/test/mocks/data/award.data.js
@@ -15,7 +15,7 @@
         'Urgent'
       ]
     },
-    case_type_category: '4',
+    case_type_category: '3',
     is_forkable: '1',
     is_forked: ''
   }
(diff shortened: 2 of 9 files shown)


ang/civiawards/shared/configs/case-type-buttons.config.js
```diff
@@ -0,0 +1,51 @@
...
+     * towards the awards configuration form.
+     *
+     * @param {object[]} caseTypes a list of case types objects.
+     */
+    function addConfigurationButtonsToCaseTypes (caseTypes) {
+      _.forEach(caseTypes, function (caseType) {
+        var caseTypeConfigUrl = url(AWARD_CONFIG_URL + caseType.id);
+
+        CaseTypeButtonsProvider.addButtons(caseType.name, [{
+          url: caseTypeConfigUrl
```
deb1990: As mentioned in the previous PR, we also need to supply the icon class from here.

ang/civiawards/shared/configs/case-type-buttons.config.js
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Adds configuration buttons to the given case types. The buttons url point
+     * towards the awards configuration form.
+     *
+     * @param {object[]} caseTypes a list of case types objects.
+     */
+    function addConfigurationButtonsToCaseTypes (caseTypes) {
+      _.forEach(caseTypes, function (caseType) {
+        var caseTypeConfigUrl = url(AWARD_CONFIG_URL + caseType.id);
```
deb1990: Also right now, the buttons would be visible for Civicase/Prospect Dashboard too right? we need to implement services for each button.

ang/civiawards/shared/configs/case-type-buttons.config.js
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    /**
+     * Returns a list of case types that belong to the given category.
+     *
+     * @param {number} categoryValue the case type category value.
+     * @returns {object[]} a list of case types.
+     */
+    function getCaseTypesForCategory (categoryValue) {
+      return _.filter(CRM['civicase-base'].caseTypes, function (caseType) {
```
deb1990: Please dont use `CRM` variable directly anymore. Lets use the Services created in `civicase-base`.

deb1990: Also lets move this function to https://github.com/compucorp/uk.co.compucorp.civicase/blob/RSE-44-awards-management/ang/civicase-base/services/case-type.service.js

ang/test/mocks/data/award.data.js
```diff
@@ -15,7 +15,7 @@
         'Urgent'
       ]
     },
-    case_type_category: '4',
+    case_type_category: '3',
```
deb1990: 4 is better :/

ang/civiawards/shared/configs/case-type-buttons.config.js
reneolivo: As discussed in Slack the icons won't be visible in other extensions because each one of them filter the case types they show on the dashboard. ie: cases only show cases case type, award dashboard only shows award case types, etc.

Since we are only adding the icons to the award case types we don't need any further filtering.

PR #21, opened 2019-11-13
CPS-25: Fix error after disabling extension

16kilobyte:
## Overview
When the extension is disabled and you try to access the url `/civicrm/admin/custom/group?reset=1`, an error is displayed as shown in the screenshot below. This PR fixes that. 
  

## Before
CPS-25-before

## After
CPS-25-after

## Technical Details
The fatal error is caused by the line `'description' => 'CRM_CiviAwards_Helper_CaseTypeCategory::getAwardCaseTypes;'` in https://github.com/compucorp/uk.co.compucorp.civiawards/blob/e9768872150931263bf1596bf3fa1b68305f8e37/CRM/CiviAwards/Setup/AddAwardsCgExtendsOptionValue.php#L30. When the module is disabled, the module files are not loaded again and when Civi tries to resolve the class  `CRM_CiviAwards_Helper_CaseTypeCategory::getAwardCaseTypes` to get the description of the Option value and fails to find the class (because it is not loaded), it throws the fatal error.

## Comments
When the extension is disabled, we remove the description and re-add it back when the extension is re-enabled

### DIFF
--- a/CRM/CiviAwards/Setup/AddAwardsCgExtendsOptionValue.php
+++ b/CRM/CiviAwards/Setup/AddAwardsCgExtendsOptionValue.php
@@ -11,6 +11,9 @@ class CRM_CiviAwards_Setup_AddAwardsCgExtendsOptionValue {
 
   /**
    * Add Awards as a valid Entity that a custom group can extend.
+   *
+   * We also update the description just incase the module was
+   * disabled before uninstallation.
    */
   public function apply() {
     $result = civicrm_api3('OptionValue', 'get', [
@@ -19,6 +22,8 @@ class CRM_CiviAwards_Setup_AddAwardsCgExtendsOptionValue {
     ]);
 
     if ($result['count'] > 0) {
+      $this->toggleOptionValueStatus(TRUE);
+
       return;
     }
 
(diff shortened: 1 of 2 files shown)


CRM/CiviAwards/Setup/AddAwardsCgExtendsOptionValue.php
```diff
@@ -19,6 +20,8 @@ public function apply() {
     ]);
 
     if ($result['count'] > 0) {
+      $this->toggle(TRUE);
```
tunbola: `$result['count'] > 0` means that the option value does not exist at all. 
No need to toggle this.

CRM/CiviAwards/Setup/AddAwardsCgExtendsOptionValue.php
```diff
@@ -33,4 +36,28 @@ public function apply() {
     ]);
   }
 
+  /**
+   * Enables/Disables the Awards option values
+   *
+   * @param $newStatus
+   *   True to enable, False to disable.
+   */
+  public function toggle($newStatus) {
```
tunbola: This function name needs to be a bit more descriptive. I would rename it to toggleStatus.
Also you need to explain in the description why toggling the status is not enough and why you also need to update the description. 
Normally setting the option value status to FALSE should suffice but there is a bug in Core, I put up am PR for it here (https://github.com/civicrm/civicrm-core/pull/15759) and when this is merged, we would not need to update the description field anymore.

CRM/CiviAwards/Setup/AddAwardsCgExtendsOptionValue.php
```diff
@@ -33,4 +36,28 @@ public function apply() {
     ]);
   }
 
+  /**
+   * Enables/Disables the Awards option values
+   *
+   * @param $newStatus
```
tunbola: Missing parameter comment.

You need to run npm install in the civiawards root so that the php linter is added to git pre-commit hook. That way you get notified of things like this before you commit.

tunbola:
@16kilobyte ,  The description in the overview is a bit too technical, Overviews are basically for non technical stuffs. You can move the tech part to the After section

Also can you add more details as to why the the line `'description' => 'CRM_CiviAwards_Helper_CaseTypeCategory::getAwardCaseTypes;'` is responsible for the error in the tech details section?

CRM/CiviAwards/Setup/AddAwardsCgExtendsOptionValue.php
```diff
@@ -33,4 +38,28 @@ public function apply() {
     ]);
   }
 
+  /**
+   * Enables/Disables the Awards option values.
```
tunbola: Can you add comment on the description as to why we are changing the description field also?

CRM/CiviAwards/Setup/AddAwardsCgExtendsOptionValue.php
```diff
@@ -19,6 +22,8 @@ public function apply() {
     ]);
 
     if ($result['count'] > 0) {
+      $this->toggle(TRUE);
```
tunbola: Can you take this line off? It's still referring to a non existent function also.

PR #27, opened 2019-11-27
RSE-616: Create Basic Details tab

deb1990:
## Overview
As part of this PR, the Basic Details Section is moved into its own tab.

## Before
2019-11-27 at 1 30 PM

## After
2019-11-27 at 1 29 PM

## Technical Details
1. Moved the markup for the Basic Details inside a tab.
2. Also "New Status" is removed.


jamienovick:
save on the right - lets look at having some standards to this please.

cancel on the left. 

Is there more detail on the ticket about these points? thanks

jamienovick:
Also this is a multiple page tabbed form. How will validation work across pages and also do I need to complete all pages before I can save? @shitijg FYI

jamienovick:
ps. Can we think about the field order here?

Type should be above description
Award manager above the start date

Width of the date fields is too wide? Is this our standard?
Enabled at the bottom please.

Please take a look at creating an event in CiviCRM

Thanks

deb1990:
> save on the right - lets look at having some standards to this please.

> cancel on the left.

> Is there more detail on the ticket about these points? thanks


> Also this is a multiple page tabbed form. How will validation work across pages and also do I need to complete all pages before I can save? @shitijg FYI

>ps. A lot of this was thought about for CiviHR so we should dig out those guidelines for UX please and follow them @deb1990 can you look that up.

Thanks @jamienovick for the feedback. These points will be discussed before starting development of RSE-597. cc @shitijg

PR #31, opened 2019-12-05
RSE-631: Sets Activity Status 'Scheduled' and 'Completed' for Applicant Review

16kilobyte:
## Overview
This PR sets correct activity status ('Completed' and 'Scheduled') options to be used by new activity category "Applicant Review"

## Before
Action was not present
RSE-631-before

## After
RSE-631-after

## Technical Details
I added a method in the upgrader to get the statuses `Completed` and `Scheduled` option values and append the name of the activity category `Applicant Review` to the grouping column.

```php
foreach ($statuses as $status) {
  civicrm_api3('OptionValue', 'create', [
    'id' => $status['id'],
    'grouping' => "{$status['grouping']},Applicant Review",
  ]);
}
```

I also added a method to undo the action above in the uninstaller.

```php
foreach ($statuses as $status) {
  $groupings = explode(',', $status['grouping']);
  $index = array_search('Applicant Review', $groupings);
  if ($index === FALSE){
    continue;
  } else {
    unset($groupings[$index]);
  }

  civicrm_api3('OptionValue', 'create', [
    'id' => $status['id'],
    'grouping' => implode(',', $groupings),
  ]);
}
```

### DIFF
--- a/CRM/CiviAwards/Setup/CreateApplicantReviewActivityType.php
+++ b/CRM/CiviAwards/Setup/CreateApplicantReviewActivityType.php
@@ -11,6 +11,7 @@ class CRM_CiviAwards_Setup_CreateApplicantReviewActivityType {
   public function apply() {
     $this->addApplicantReviewCategory();
     $this->addApplicantReviewActivityType();
+    $this->setApplicantReviewActivityTypeStatuses();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


CRM/CiviAwards/Setup/CreateApplicantReviewActivityType.php
```diff
@@ -45,4 +46,23 @@ private function addApplicantReviewActivityType() {
     ]);
   }
 
+  /**
+   * Sets activity status 'Scheduled' and 'Completed' for applicant
+   * review activity type.
+   */
+  private function setApplicantReviewActivityTypeStatuses() {
+    $statuses = civicrm_api3('OptionValue', 'get', [
+      'sequential' => 1,
```
tunbola: No need for the sequential parameter here.

CRM/CiviAwards/Setup/DeleteApplicantReviewCustomField.php
```diff
@@ -6,9 +6,17 @@
 class CRM_CiviAwards_Setup_DeleteApplicantReviewCustomField {
 
   /**
-   * Deletes the Applicant review custom group set.
+   * Removes the Applicant review activity group and statuses.
    */
   public function apply() {
+    $this->removeApplicantReviewCustomGroup();
```
tunbola: I don't think we need the cleanup part for the Applicant review statuses since we are not deleting the applicant review activity type. Besides this is also a wrong place to add logic like this since this class names describes that it deletes custom fields.

Here is what I will suggest:

- Let's revert this file back to what it was.
- Let's make the setApplicantReviewActivityTypeStatuses in the `CRM_CiviAwards_Setup_CreateApplicantReviewActivityType` to be idempotent, i.e we check if the `Applicant Review ` is already part of the grouping for the concerned statuses, if it is, we don't bother adding it. 

That way if the extension is un-installed and re-installed, we won't have duplicate data

CRM/CiviAwards/Setup/CreateApplicantReviewActivityType.php
```diff
@@ -45,4 +46,28 @@ private function addApplicantReviewActivityType() {
...
+    foreach ($statuses as $status) {
+      $groupings = explode(',', $status['grouping']);
+      $index = array_search('Applicant Review', $groupings);
+      if ($index !== FALSE){
+        continue;
+      }
+
+      civicrm_api3('OptionValue', 'create', [
+        'id' => $status['id'],
+        'grouping' => "{$status['grouping']},Applicant Review",
```
tunbola: What is {$status['grouping']} is empty or NULL? In that case we won't need the comma before applicant review

PR #33, opened 2019-12-06
RSE-657: Add Create Edit Awards Permissions

16kilobyte:
## Overview
This PR adds `create/edit` permission for the extension

## Before
Permission did not exists

## After
RSE-657-after

### DIFF
--- a/civiawards.php
+++ b/civiawards.php
@@ -161,6 +161,11 @@ function civiawards_civicrm_permission(&$permissions) {
     'CiviAwards: Access applicant portal',
     ts('Allows a user to access the awards applicant portal'),
   ];
+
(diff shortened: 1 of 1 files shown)


civiawards.php
```diff
@@ -161,6 +161,11 @@ function civiawards_civicrm_permission(&$permissions) {
     'CiviAwards: Access applicant portal',
     ts('Allows a user to access the awards applicant portal'),
   ];
+
+  $permissions['create/edit awards'] = [
+    'CiviAwards: Create/edit awards',
```
tunbola: Can we change the label to `CiviAwards: Create/Edit awards` i.e make edit to be sentence case.

PR #34, opened 2019-12-09
RSE-372: Save More Information for Each Review Field

16kilobyte:
## Overview
At the moment, only the ID of every review field is saved as part of “AwardDetails“ API.

Now we need support to add “weight“, “required“ fields too.

 

Also there is a bug with existing implementation of “review_fields“. it always returns an extra NULL value. If an award is saved with [“19“, 20”], then the api return [“19”, “20“, ““].

## Before
Action was not preset
RSE-372-before

## After
RSE-372-after
RSE-372-after-2

## Technical Details
In the method `createProfileFields` , Instead of expecting just a dimensional array of custom field id, the method now expects a 2-dimensional array in the form `[['id' => 1, 'required' => FALSE, 'weight' => 3]]`. This field will now be used to create the `UFField` entry. 

To show, in the method `getProfileFields` , instead of just iterating over the custom fields and stripping and returning the IDs:
```php
foreach ($result['values'] as $profileField) {
  $customFields[] = [
    'id' => substr($profileField['field_name'], 7),
    'required' => $profileField['is_required'],
    'weight' => $profileField['weight'],
  ];
}
```

### DIFF
--- a/CRM/CiviAwards/BAO/AwardDetail.php
+++ b/CRM/CiviAwards/BAO/AwardDetail.php
@@ -97,7 +97,6 @@ class CRM_CiviAwards_BAO_AwardDetail extends CRM_CiviAwards_DAO_AwardDetail {
       return;
     }
 
-    $filteredReviewFields = [];
     foreach ($reviewFields as $reviewFieldOptions) {
       $invalidReviewFieldOptions = [];
       $validReviewFieldOptions = array_filter(
@@ -133,11 +132,7 @@ class CRM_CiviAwards_BAO_AwardDetail extends CRM_CiviAwards_DAO_AwardDetail {
         1 => implode(', ', $invalidReviewFieldOptions),
         2 => implode(', ', $missingReviewFieldOptions),
       ]));
-
-      $filteredReviewFields[] = $validReviewFieldOptions;
     }
-
-    $params['review_fields'] = $filteredReviewFields;
   }
 
 }
--- a/CRM/CiviAwards/BAO/AwardDetail.php
+++ b/CRM/CiviAwards/BAO/AwardDetail.php
@@ -86,53 +86,10 @@ class CRM_CiviAwards_BAO_AwardDetail extends CRM_CiviAwards_DAO_AwardDetail {
    *
    * @throws \Exception
    */
-  private static function validateReviewFields(array &$params) {
-    $reviewFieldsLegalOptions = [
-      'id' => FILTER_VALIDATE_INT,
-      'weight' => FILTER_VALIDATE_INT,
-      'required' => FILTER_VALIDATE_BOOLEAN
-    ];
+  private static function validateReviewFields(array $params) {
     $reviewFields = CRM_Utils_Array::value('review_fields', $params);
-    if (empty($reviewFields)) {
-      return;
-    }
-
-    foreach ($reviewFields as $reviewFieldOptions) {
-      $invalidReviewFieldOptions = [];
-      $validReviewFieldOptions = array_filter(
-        $reviewFieldOptions,
-        function($value, $key) use ($reviewFieldsLegalOptions, &$invalidReviewFieldOptions) {
-          if (!isset($reviewFieldsLegalOptions[$key])) {
-            return FALSE;
-          }
-
-          if (filter_var($value, $reviewFieldsLegalOptions[$key], FILTER_NULL_ON_FAILURE) !== NULL) {
-            return TRUE;
-          }
-          $invalidReviewFieldOptions[] = $key;
-        },
-        ARRAY_FILTER_USE_BOTH
-      );
-      if (count($validReviewFieldOptions) === count($reviewFieldsLegalOptions)) {
-        continue;
-      }
-
-      $missingReviewFieldOptions = array_diff(
-        array_keys($reviewFieldsLegalOptions),
-        array_merge($invalidReviewFieldOptions, array_keys($validReviewFieldOptions))
-      );
-      $exceptionPhrases = [
-        !empty($invalidReviewFieldOptions) ? "invalid values passed for [%1]" : "",
-        !empty($missingReviewFieldOptions) ? "required field" . (
-          count($missingReviewFieldOptions) > 1 ? "s [%2] are" : " [%2] is"
-        ) . " missing" : ""
-      ];
-
-      throw new Exception(ts(implode(' and ', array_filter($exceptionPhrases)), [
-        1 => implode(', ', $invalidReviewFieldOptions),
-        2 => implode(', ', $missingReviewFieldOptions),
-      ]));
-    }
+    $reviewFieldsValidator = new CRM_CiviAwards_Helper_CreateReviewFieldsValidator();
+    $reviewFieldsValidator->validate($reviewFields);
   }
 
 }
--- a/CRM/CiviAwards/BAO/AwardDetail.php
+++ b/CRM/CiviAwards/BAO/AwardDetail.php
@@ -55,8 +55,9 @@ class CRM_CiviAwards_BAO_AwardDetail extends CRM_CiviAwards_DAO_AwardDetail {
    * @param array $params
    *   Parameters.
    */
-  private static function validateParams(array $params) {
+  private static function validateParams(array &$params) {
     self::validateDates($params);
+    self::validateReviewFields($params);
   }
 
   /**
@@ -74,6 +75,69 @@ class CRM_CiviAwards_BAO_AwardDetail extends CRM_CiviAwards_DAO_AwardDetail {
         throw new Exception("Award Start Date must not be greater than the End Date");
       }
     }
+  }
+
+
+  /**
+   * Validates review_field options to ensure legal options are passed.
+   *
+   * @param array $params
+   *  Request Parameters
+   *
+   * @throws \Exception
+   */
+  private static function validateReviewFields(array &$params) {
+    $reviewFieldsLegalOptions = [
+      'id' => FILTER_VALIDATE_INT,
+      'weight' => FILTER_VALIDATE_INT,
+      'required' => FILTER_VALIDATE_BOOLEAN
+    ];
+    $reviewFields = CRM_Utils_Array::value('review_fields', $params);
+    if (empty($reviewFields)) {
+      return;
+    }
+
+    $filteredReviewFields = [];
+    foreach ($reviewFields as $reviewFieldOptions) {
+      $invalidReviewFieldOptions = [];
+      $validReviewFieldOptions = array_filter(
+        $reviewFieldOptions,
+        function($value, $key) use ($reviewFieldsLegalOptions, &$invalidReviewFieldOptions) {
+          if (!isset($reviewFieldsLegalOptions[$key])) {
+            return FALSE;
+          }
+
+          if (filter_var($value, $reviewFieldsLegalOptions[$key], FILTER_NULL_ON_FAILURE) !== NULL) {
+            return TRUE;
+          }
+          $invalidReviewFieldOptions[] = $key;
+        },
+        ARRAY_FILTER_USE_BOTH
+      );
+      if (count($validReviewFieldOptions) === count($reviewFieldsLegalOptions)) {
+        continue;
+      }
+
+      $missingReviewFieldOptions = array_diff(
+        array_keys($reviewFieldsLegalOptions),
+        array_merge($invalidReviewFieldOptions, array_keys($validReviewFieldOptions))
+      );
+      $exceptionPhrases = [
+        !empty($invalidReviewFieldOptions) ? "invalid values passed for [%1]" : "",
+        !empty($missingReviewFieldOptions) ? "required field" . (
+          count($missingReviewFieldOptions) > 1 ? "s [%2] are" : " [%2] is"
+        ) . " missing" : ""
+      ];
+
+      throw new Exception(ts(implode(' and ', array_filter($exceptionPhrases)), [
+        1 => implode(', ', $invalidReviewFieldOptions),
+        2 => implode(', ', $missingReviewFieldOptions),
+      ]));
+
+      $filteredReviewFields[] = $validReviewFieldOptions;
+    }
+
+    $params['review_fields'] = $filteredReviewFields;
   }
 
 }
--- a/CRM/CiviAwards/Helper/CreateReviewFieldsValidator.php
+++ b/CRM/CiviAwards/Helper/CreateReviewFieldsValidator.php
@@ -6,25 +6,25 @@
 class CRM_CiviAwards_Helper_CreateReviewFieldsValidator {
 
   /**
-   * Array of accepted review_fields with their validation filters
+   * Array of accepted review_fields with their validation filters.
    *
    * @var array
    */
   private $reviewFieldsLegalOptions = [
     'id' => FILTER_VALIDATE_INT,
     'weight' => FILTER_VALIDATE_INT,
-    'required' => FILTER_VALIDATE_BOOLEAN
+    'required' => FILTER_VALIDATE_BOOLEAN,
   ];
 
   /**
-   * Array to hold invalid field options
+   * Array to hold invalid field options.
    *
    * @var array
    */
   private $invalidReviewFieldOptions = [];
 
   /**
-   * Array to hold unknown review field options passeed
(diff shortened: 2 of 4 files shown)


CRM/CiviAwards/Service/AwardProfile.php
```diff
@@ -28,20 +28,30 @@ public function createProfile($caseTypeId) {
...
+    foreach ($customFields as $customField) {
       civicrm_api3('UFField', 'create', [
         'uf_group_id' => $profileId,
-        'field_name' => 'custom_' . $customFieldId,
+        'field_name' => 'custom_' . $customField['id'],
         'field_type' => 'Activity',
-        'label' => 'Activity Field' . $customFieldId,
+        'label' => 'Activity Field' . $customField['id'],
+        'is_required' => boolval($customField['required']),
+        // Nesting api call because of a bug that seems to be if you have two profile fields
```
tunbola: If you can refine the comment here. See https://github.com/compucorp/uk.co.compucorp.civicrm.prospect/blob/master/CRM/Prospect/Setup/CreateProspectMenus.php#L48-L49 also. Just say that civi will assign the next available weight when creating a profile field irrespective of the value passed in the weight field.

Might not be a bug but how civi works with weight fields on create. Might be worth taking it up with the community one of these days but it seems to be a minor issue.

CRM/CiviAwards/Service/AwardProfile.php
```diff
@@ -28,20 +28,30 @@ public function createProfile($caseTypeId) {
...
+        // in a group and you are about creating another field and you set the weight to say 5.
+        // Civicrm will automatically set it to 3 so that the weight is sequential
+        'api.UFField.create' => [
+          'id' => '$value.id',
+          'weight' => $customField['weight'],
+        ],
       ]);
     }
+    $this->addContactIdField($profileId);
```
tunbola: I don't think we need this `$this->addContactIdField($profileId);` It is already handled outside the loop.

CRM/CiviAwards/Service/AwardProfile.php
```diff
@@ -53,7 +63,25 @@ public function createProfileFields(array $customFields, $profileId) {
     $this->addContactIdField($profileId);
   }
 
-    $this->addContactIdField($profileId);
+  /**
+   * Validates review_field options to ensure legal options are passed.
+   *
+   * @param array
```
tunbola: No parameter name?
Did you run the git pre-commit hook on local before committing?

CRM/CiviAwards/Service/AwardProfile.php
```diff
@@ -53,7 +63,25 @@ public function createProfileFields(array $customFields, $profileId) {
...
+   *  Custom field
+   *
+   * @throws \CiviCRM_API3_Exception
+   */
+  private function validateReviewFields($customField) {
```
tunbola: No function type hinting for $customField? 
Please make sure to run the npm install and the pre-commit hook. Things like this are not supposed to be brought up in code review because the linters should have taken care of that already.

Thanks

CRM/CiviAwards/Service/AwardProfile.php
```diff
@@ -53,7 +63,25 @@ public function createProfileFields(array $customFields, $profileId) {
...
+    $missingFieldsDiff = array_diff($this->reviewFieldsLegalOptions, array_keys($customField));
+    if (count($missingFieldsDiff) > 0) {
+      $missingFields = implode(', ', $missingFieldsDiff);
+      $exceptionPhrase = (
+        count($missingFieldsDiff) > 1 ? "fields [%1] are" : "field [%1] is"
+      );
+      throw new CiviCRM_API3_Exception(ts("Required $exceptionPhrase missing in review_fields", [
```
tunbola: Can we just throw a normal exception here and not an API exception?
The reason is that the Award profile class may be used internally in code and not just by the API.

CRM/CiviAwards/Service/AwardProfile.php
tunbola: Also, we need to move the validation logic to the AwardDetails BAO. See: https://github.com/compucorp/uk.co.compucorp.civiawards/blob/RSE-44-awards-management/CRM/CiviAwards/BAO/AwardDetail.php#L59
The reason is because
(1)  it's better to have all the validations for an Award details in one place
(2) If we have it here, it is possible to create an Award detail successfully and the Award Profile creation will fail because of a validation error since this is an after save of the Award detail creation.

CRM/CiviAwards/BAO/AwardDetail.php
```diff
@@ -76,4 +77,34 @@ private static function validateDates(array $params) {
...
+   * @param array $params
+   *  Request Parameters
+   *
+   * @throws \Exception
+   */
+  private static function validateReviewFields(array $params) {
+    $reviewFieldsLegalOptions = ['id', 'weight', 'required'];
+    $reviewFieldsOptions = CRM_Utils_Array::value('review_fields', $params);
+    if (!$reviewFieldsOptions) {
+      throw new Exception(ts('Valid review fields options must be set'));
```
tunbola: We should not throw an exception if the review fields are not sent. The Award details UI is structured in such a way that basic fields are added on create and then, other fields can be added on edit. If you see here: https://github.com/compucorp/uk.co.compucorp.civiawards/blob/RSE-44-awards-management/CRM/CiviAwards/Service/AwardDetail.php#L164, it shows that the review fields can actually be empty.

CRM/CiviAwards/BAO/AwardDetail.php
```diff
@@ -76,4 +77,34 @@ private static function validateDates(array $params) {
...
+    }
+
+    foreach ($reviewFieldsOptions as $reviewFieldsOption) {
+      $missingFieldsDiff = array_diff($reviewFieldsLegalOptions, array_keys($reviewFieldsOption));
```
tunbola: Maybe we can go a little step further and check that the values of these fields are what we expect.
We can restructure the reviewFieldsLegalOption array and use the https://www.php.net/manual/en/function.filter-var.php function to do the validation
```
  $reviewFieldsLegalOptions = [
    [
      'name' => 'id',
      'validation' => FILTER_VALIDATE_INT
    ],
    
    [
    'name' => 'weight',
    'validation' => FILTER_VALIDATE_INT
    ]
  ];
```

CRM/CiviAwards/BAO/AwardDetail.php
```diff
@@ -76,4 +77,42 @@ private static function validateDates(array $params) {
...
+   * @throws \Exception
+   */
+  private static function validateReviewFields(array &$params) {
+    $reviewFieldsLegalOptions = [
+      'id' => FILTER_VALIDATE_INT,
+      'weight' => FILTER_VALIDATE_INT,
+      'required' => FILTER_VALIDATE_BOOLEAN
+    ];
+    $reviewFields = CRM_Utils_Array::value('review_fields', $params);
+    if (is_array($reviewFields)) {
```
tunbola: Let's do a quick return here instead of nesting multiple if's

i.e 
```php
if (empty ($reviewFields)) {
return;
}
```

CRM/CiviAwards/BAO/AwardDetail.php
```diff
@@ -76,4 +77,42 @@ private static function validateDates(array $params) {
...
+      foreach ($reviewFields as &$reviewFieldOptions) {
```
tunbola: Do we need to pass by reference here?

CRM/CiviAwards/BAO/AwardDetail.php
```diff
@@ -76,4 +77,42 @@ private static function validateDates(array $params) {
...
+        $reviewFieldOptions = filter_var_array($reviewFieldOptions, $reviewFieldsLegalOptions, FALSE);
```
tunbola: Are we supposed to iterate over the `$reviewFieldOptions` at all? 
Looking at the manual here: https://www.php.net/manual/en/function.filter-var-array.php
The first array will contain the key => value pairs for the review fields but here the review fields are being iterated and the key is not even present.

CRM/CiviAwards/BAO/AwardDetail.php
```diff
@@ -76,4 +77,42 @@ private static function validateDates(array $params) {
...
+        if (count($reviewFieldOptions) !== count($reviewFieldsLegalOptions)) {
```
tunbola: Also if using the filter_var_array function, if a field fails validation, the field is still present, the value will be false. See: https://3v4l.org/nk4PS

CRM/CiviAwards/BAO/AwardDetail.php
16kilobyte: I want to modify the array itself with filtered values instead making a new copy of the array again

CRM/CiviAwards/BAO/AwardDetail.php
16kilobyte: The thing is the `$reviewFields` come as `[['id' => 34, 'required' => true, 'weight' => 5], ['id' => 45, 'required' => false, 'weight' => 14]]`

CRM/CiviAwards/BAO/AwardDetail.php
```diff
@@ -76,4 +77,67 @@ private static function validateDates(array $params) {
...
+    if (empty($reviewFields)) {
+      return;
+    }
+
+    $filteredReviewFields = [];
+    foreach ($reviewFields as $reviewFieldOptions) {
+      $invalidReviewFieldOptions = [];
+      $validReviewFieldOptions = array_filter(
+        $reviewFieldOptions,
+        function($value, $key) use ($reviewFieldsLegalOptions, &$invalidReviewFieldOptions) {
```
tunbola: Can we move this to a separate function rather and use he function name in the array_filter here. Makes the code look less cluttered.

CRM/CiviAwards/BAO/AwardDetail.php
```diff
@@ -76,4 +77,67 @@ private static function validateDates(array $params) {
...
+          if (!isset($reviewFieldsLegalOptions[$key])) {
```
tunbola: We should probably store these fields in an array too because they are the `illegal` field options

CRM/CiviAwards/BAO/AwardDetail.php
```diff
@@ -76,4 +77,67 @@ private static function validateDates(array $params) {
...
+            return FALSE;
+          }
+
+          if (filter_var($value, $reviewFieldsLegalOptions[$key], FILTER_NULL_ON_FAILURE) !== NULL) {
```
tunbola: I have some reservations about suing the `FILTER_NULL_ON_FAILURE` flag because if you check the manual here: https://www.php.net/manual/en/filter.filters.validate.php, the filer is valid for `FILTER_VALIDATE_BOOLEAN` only and not for `FILTER_VALIDATE_INT`, so we should find an alternative.

CRM/CiviAwards/BAO/AwardDetail.php
```diff
@@ -76,4 +77,67 @@ private static function validateDates(array $params) {
...
+          count($missingReviewFieldOptions) > 1 ? "s [%2] are" : " [%2] is"
+        ) . " missing" : ""
+      ];
+
+      throw new Exception(ts(implode(' and ', array_filter($exceptionPhrases)), [
+        1 => implode(', ', $invalidReviewFieldOptions),
+        2 => implode(', ', $missingReviewFieldOptions),
+      ]));
+
+      $filteredReviewFields[] = $validReviewFieldOptions;
```
tunbola: I don't think this is necessary. Also the code will never get to this stage because once the review fields are valid, the code will never get to this place and the loop continues. 
```php
     if (count($validReviewFieldOptions) === count($reviewFieldsLegalOptions)) {
        continue;
      }
```
Then, exceptions will definitely be thrown, so code will not get to this stage.
So if the review fields are valid, technically there is a bug here because `filteredReviewFields` will be empty and the line below `$params['review_fields'] = $filteredReviewFields;` will set the review fields to empty.

CRM/CiviAwards/BAO/AwardDetail.php
```diff
@@ -76,4 +77,67 @@ private static function validateDates(array $params) {
...
+    }
+
+    $params['review_fields'] = $filteredReviewFields;
```
tunbola: No need for this line.

CRM/CiviAwards/BAO/AwardDetail.php
16kilobyte: You are right. I have removed this bit

CRM/CiviAwards/Helper/CreateReviewFieldsValidator.php
```diff
@@ -0,0 +1,109 @@
...
+   * Callback to filter review options.
+   *
+   * @param string $value
+   *  Value of review field option
+   * @param string $key
+   *  Review field option
+   *
+   * @return boolean
+   */
+  private function reviewFieldOptionsFilter(string $value, string $key) {
```
tunbola: We usually don't do typehinting for scalar types. Let's remove them for consistency.

CRM/CiviAwards/Helper/CreateReviewFieldsValidator.php
```diff
@@ -0,0 +1,109 @@
...
+
+  /**
+   * Callback to filter review options.
+   *
+   * @param string $value
+   *  Value of review field option
+   * @param string $key
+   *  Review field option
+   *
+   * @return boolean
```
tunbola: I'm very sure you didn't run the linters because this is supposed to have a description/comment if you did. Can you please do? Thanks.

CRM/CiviAwards/Helper/CreateReviewFieldsValidator.php
```diff
@@ -0,0 +1,109 @@
...
+   * @throws \Exception
+   */
+  public function validate(array $reviewFields) {
+    if (empty($reviewFields)) {
+      return;
+    }
+
+    foreach ($reviewFields as $reviewFieldOptions) {
+      $validReviewFieldOptions = array_filter($reviewFieldOptions, [
+        $this,
```
tunbola: Can you please make sure what is passed as the second parameter is on a single line
```php
[$this, 'reviewFieldOptionsFilter']
```
Thanks.

CRM/CiviAwards/Helper/CreateReviewFieldsValidator.php
```diff
@@ -0,0 +1,109 @@
...
+   *  Value of review field option
+   * @param string $key
+   *  Review field option
+   *
+   * @return boolean
+   */
+  private function reviewFieldOptionsFilter(string $value, string $key) {
+    
+    if (!isset($this->reviewFieldsLegalOptions[$key])) {
+      return FALSE;
```
tunbola: I mentioned earlier that we should store the value here in an array. These are actually illegal fields, i.e fields added to the review fields that is not part of the fields we are expecting.

CRM/CiviAwards/Helper/CreateReviewFieldsValidator.php
```diff
@@ -0,0 +1,109 @@
+<?php
+
+/**
+ * Class CRM_CiviAwards_Helper_CreateReviewFieldsValidator.
+ */
+class CRM_CiviAwards_Helper_CreateReviewFieldsValidator {
```
tunbola: Since you have decided to move this to a class, it would have been better to make this generic a bit such that the constructor will accept `$fieldsLegalOptions` array and this could be used to validate arrays having similar structure. But let's do this later some other time since this has taken some time already.

CRM/CiviAwards/Helper/CreateReviewFieldsValidator.php
```diff
@@ -0,0 +1,124 @@
...
+   *
+   * @var array
+   */
+  private $unknownReviewFieldOptions = [];
+
+  /**
+   * Validates review_field options to ensure legal options are passed.
+   *
+   * @param array $reviewFields
+   *  Review fields
```
tunbola: Can you PLEASE make sure you run the linters locally. Tried committing this locally and got 15 errors.
This purpose of running it is to make the PR review process smoother and make it easier for the reviewer. If we commit this file this way, the next person working on it will have to deal with fixing the errors you could have fixed.

PR #36, opened 2019-12-10
RSE-668: Enable AwardManager API

16kilobyte:
## Overview
This PR adds an API file for `AwardManager` entity to enable API calls

## Before
API actions does not exist for the `AwardManager` entity

## After
RSE-after

### DIFF
--- a/api/v3/AwardManager.php
+++ b/api/v3/AwardManager.php
@@ -1,0 +1,47 @@
+<?php
+
+/**
+ * @file
+ * Award Manager API file.
+ */
+
+/**
+ * AwardManager.create API.
+ *
+ * @param array $params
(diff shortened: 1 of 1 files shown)


api/v3/AwardManager.php
```diff
@@ -0,0 +1,67 @@
...
+ * Award Manager API file.
+ */
+
+/**
+ * AwardManager.create API specification.
+ *
+ * @param array $spec
+ *   Description of fields supported by this API call.
+ */
+function _civicrm_api3_award_manager_create_spec(array &$spec) {
```
tunbola: No need to add the spec function since we are not adding any pseudo fields. Civi should build the parameters itself.

PR #39, opened 2019-12-18
RSE-681: Order Profile Fields by Weight on Review Form

16kilobyte:
## Overview
The award review fields are now created with more properties like weight, required.
On the Review form which is basically the profile containing the chosen custom fields for the Award, we need to make sure that the fields are ordered by the weight.

## Before
No logic to ensure fields were ordered by their weights

## After
Profile fields are now ordered with `weight`
RSE-681

## Technical Details
We just set the default options for `CRM_Core_BAO_UFGroup::getFields()` and set `weight` as the `$orderBy` param which is the 11th parameter


tunbola:
@16kilobyte , The before section image is not really useful as it just shows a a table with no header fields, if you can add a proper one or remove it entirely.

PR #44, opened 2019-12-29
RSE-359: Reviews tab

reneolivo:
## Overview
This PR adds the Reviews tab to the case details.

**Note:** This PR is related to https://github.com/compucorp/uk.co.compucorp.civicase/pull/317

## Before
Screen Shot 2019-12-29 at 6 20 22 PM

## After
pr-after

## Technical Details

A new review case tab service was created so we can add reviews tab placeholder elements, and the directive to display the actual review tab.

### Scoring fields

Each application must be reviewed using the scoring fields as defined in the award details configuration.

When the review tab loads it starts loading the award details (case type) because they hold the custom fields used for reviewing. It also loads all the custom fields related to award's application reviews because the award details only have the name, and the sort order, but not the label for the field.

### Creating and editing reviews

When creating or editing reviews we use the `crmFormSuccess.crmPopup crmPopupFormSuccess.crmPopup` event instead of `crmFormSuccess` or the `crm-form-success` directive because the form does not trigger the `crmFormSuccess` since it never closes after creating a review. It redirects to view the newly created review. The even for submitting the form successfully is the one used on this PR.


jamienovick:
@reneolivo looks good - nice steps and I like the fact that we can add new tabs as this is something we will want to do in future :)

With respect to styling:

1. Can the table in:

https://user-images.githubusercontent.com/1642119/71563392-05777600-2a65-11ea-8ac2-dda61db8c1ec.gif

Have a shadow so that it's not flat on the page i.e. like a panel? 

2. I see that we are including the review fields. This was actually removed from the spec @shitijg @jerry-compucorp but since its done that is great. 

How will the system deal with having a large number of fields however? For example if there are 10 or 15? Can the table scroll fixing the non scoring fields?

reneolivo:
@jamienovick the review fields were part of the spec before it was updated. I think this was done before the update. Not sure how long the review fields form took, but displaying them on this table was pretty simple.

Regarding the number of columns, that's a valid concern. I could try to fix the columns. Need to check with @deb1990 and @swastikpareek since I gave it a quick try and was not able to do it. Maybe they have experience in that regard.

Another solution might be to display a `...` column if there are more than 3 review fields. If the `...` is clicked it can open the `view review` popup.

shitijg:
@jamienovick 
We have a couple of options here:
- We implement horizontal scroll with some columns frozen. We have done this before with Manage Cases screen and this is a bit complex to implement (cc: @reneolivo @deb1990 )

- Just create a scrollable table. However, the actions column will be at the end of the scroll which is not ideal (maybe have actions as the first column? or make the whole row clickable and actions available with a right click - not preferred)  

- As @reneolivo suggested, as it stands, the user can still click on view and see all review fields. From what I understand - the objective here is to see all the review fields in one go (and tabular format is preferred). We can add an export as CSV functionality which might be easier? And / Or have a functionality for the admin user to select which three columns to see on the tabular view 

Let us know! 

Thanks

jamienovick:
Hi guys, the options are no help without estimates.

**1. We implement horizontal scroll with some columns frozen. We have done this before with Manage Cases screen and this is a bit complex to implement (cc: @reneolivo @deb1990 )**

What is the estimate? i.e. https://ant.design/components/table/#components-table-demo-fixed-columns-header

2. Just create a scrollable table. However, the actions column will be at the end of the scroll which is not ideal (maybe have actions as the first column? or make the whole row clickable and actions available with a right click - not preferred)I assume you mean that there would be no fixed first of last row. This would not be good.

3. As @reneolivo suggested, as it stands, the user can still click on view and see all review fields. From what I understand - the objective here is to see all the review fields in one go (and tabular format is preferred). We can add an export as CSV functionality which might be easier? And / Or have a functionality for the admin user to select which three columns to see on the tabular viewYou have export as CSV as its activities (from the search activities) so this is not part of the requirement.

Other options:

4. Have a property of the fields to "show in table"

We can add a property to the awards fields in the award configuration to show in table. We could use an appropriate field from the profile entity to store the data about this functionality (i.e. in_selector or in_Search).

5. Remove the fields:

The spec was modified to exclude these fields for the exact reason above. Given that we've spent some time on this, lets explore 

My proposed preference order is:
4 - 1 - 5 depending on the estimates. If the estimate for 4 or 1 is <8h then let's do that.

Thanks

reneolivo:
@jamienovick @shitijg 

Estimates for solutions:

* Option 1: 2 to 3 days.
* Option 4: around 4 hours to 6 hours.
* Option 5: 30 mins or less.

Option 6: We were also discussing an extra option which is to have a field right on the reviews tab where the user can select which review fields to display. We could use a field similar to the tags one:
gif
I believe this option would take around 2 to 3 hours to implement. I feel it's a better solution than Option 4.

jamienovick:
@reneolivo  option 6 sounds good. 

I guess the question I would ask is how does it look when you have too many fields? In what way does the screen end up breaking?

deb1990:
If we have a popup instead of a dropdown. just like we did with selection of review fields in Create Award, it wont break the page.

jamienovick:
@reneolivo @deb1990 Sorry! Your right actually there are 2 different questions:

1. The select fields multiselect could get very long if we have say 20 fields and select them all and this would not be ideal. As such if we do have a fields select it should be in a modal popup rather then on the screen.
2. My question above was actually "how does the table look with a lot of fields". I've not seen it with 20 fields. Is it so bad.

Subject to seeing point 2, on reflection I think we should just go ahead and hide the fields from the table for now and we can add this in a future version...

J

reneolivo:
@jamienovick repeated the same fields over and over. Here's the result:

Right now we don't have a scroller, so if it's a big table the content will be hidden outright:
Screen Shot 2020-01-09 at 11 41 03 AM

If we add a scroll element then it looks like this:
gif

jamienovick:
@reneolivo lets just let it have scroll like that but can we make sure the action menu is outside the scroll?

reneolivo:
@jamienovick making it scroll is very easy, but fixing an element takes time. The problem is that to fix the element we need to use javascript to create a new element outside of the scrollable table. The other issue is that different rows might have different heights, but since the actions column does not belong to the same table we need to try and match this height using Javascript as well.

I can try and and accomplish it by timeboxing it for 4 hours. Do you think that's ok?

reneolivo:
How the reviews tab looks now with shadows and horizontal scroll:
gif

ang/civiawards/reviews-tab/directives/reviews-case-tab-content.html
```diff
@@ -41,14 +41,14 @@
               dropdown-append-to="'#bootstrap-theme'"
             >
               <button
-                class="btn btn-default btn-sm dropdown-toggle"
+                class="btn btn-default btn-sm dropdown-toggle civiawards__reviews-tab__record-action-button"
```
jamienovick: @reneolivo can't we make a generic button for above a panel with a shadow etc? That would be standard.

PR #46, opened 2019-12-31
RSE-404: "More Filters" button on Award Dashboard

deb1990:
## Overview
As part of this PR, the "More Filters" button and the separator is added to the Dashboard Action Bar.

## Before
2020-01-06 at 12 57 PM

## After
after

## Technical Details
1. `awards-dashboard-action-buttons.config.js` is renamed to `awards-dashboard-action-items.config.js`. Also a `separator` and `More Filter` button is added.
2. `add-award-dashboard-action-button` partial and controller added.
3. `dashboard-action-separator` partial and controller added.
4. `more-filters-dashboard-action-button` partial and controller added.


ang/civiawards/dashboard/directives/dashboard-action-separator.js
```diff
@@ -0,0 +1,26 @@
...
+     *
+     * @returns {boolean} true when case type category url param is awards
+     */
+    function isVisible () {
+      var urlParams = $location.search();
+
+      return urlParams.case_type_category === 'awards';
+    }
+  }
+})(angular, CRM._);
```
reneolivo: I think we need to change the visibility approach here to avoid implementing the same logic for multiple buttons.

- `awards-dashboard-action-items.config.js` should add these buttons conditionally only if the `case_type_category` param is equal to `awards`.
- `dashboard-action-separator.html` should have no `ng-controller` or `ng-show`.
- Delete `dashboard-action-separator.js`.
- Remove the `isVisible` method from `more-filters-dashboard-action-button.controller.js` and `add-award-dashboard-action-button.controller.js`

ang/civiawards/dashboard/directives/add-award-dashboard-action-button.controller.js
```diff
@@ -1,17 +1,18 @@
...
-   * @param {object} $window the window object reference.
+   * @param {object} $scope the scope object
+   * @param {object} $location the location service
+   * @param {object} $window the window object reference
    */
-  function AddAwardDashboardActionButton ($location, $window) {
-    this.clickHandler = clickHandler;
-    this.isVisible = isVisible;
+  function AddAwardDashboardActionButtonController ($scope, $location, $window) {
+    $scope.clickHandler = clickHandler;
```
reneolivo: since this is not a generic method any longer we should rename it to a more specific `redirectToNewAwardForm` or similar.

ang/civiawards/dashboard/directives/more-filters-dashboard-action-button.controller.js
```diff
@@ -0,0 +1,171 @@
...
+        awardFilter: 'my_awards',
+        statuses: '',
+        award_types: '',
+        start_date: null,
+        end_date: null
+      },
+      applyFilterAndCloseDialog: applyFilterAndCloseDialog
+    };
+
+    $scope.clickHandler = clickHandler;
```
reneolivo: This should be `openMoreFiltersDialog` or similar.

ang/civiawards/dashboard/directives/more-filters-dashboard-action-button.controller.js
```diff
@@ -0,0 +1,171 @@
...
+    function applyFilterAndCloseDialog () {
+      applyFilter();
+      dialogService.close('MoreFilters');
+    }
+    /**
+     *
+     */
+    function applyFilter () {
+      processMyAwardsFilter()
+        .then(processAwardTypeFilters)
```
reneolivo: I think these two can run at the same time so it should be:

```js
$q.all({
  myAwards: processMyAwardsFilter(),
  awardType: processAwardTypeFilters()
})
  .then(function (results) {
    $rootScope.$broadcast('civicase::dashboard-filters::updated', {
      case_type_id: { IN: results.awardType }
    });
  });
```

But, there are two things to consider:

1 - why do `processMyAwardsFilter` and `processAwardTypeFilters` both return a list of awards ids but we only care about the ones returned by `processAwardTypeFilters` ?
2 - can we try to use `applications` and `awards` instead of `awards` and `awardTypes` to avoid confusion? The former are the proper terms for cases and case types.

ang/civiawards/dashboard/directives/more-filters-dashboard-action-button.controller.js
```diff
@@ -0,0 +1,171 @@
...
+          return awardsData.values.map(function (award) {
+            return award.case_type_id;
+          });
+        });
+    }
+
+    /**
+     * Process My Awards/All Awards filters
+     *
+     * @returns {Promise} promise
```
reneolivo: should specify the resolve value. Can be `@returns {Promise<number[]>}` a promise that resolves to a list of award ids`.

ang/civiawards/dashboard/directives/more-filters-dashboard-action-button.controller.js
```diff
@@ -0,0 +1,171 @@
...
+            case_type_id: { IN: awardTypeIds }
+          });
+        });
+    }
+
+    /**
+     * Process Award Type Filters
+     *
+     * @param {number[]} caseTypeIDs case type ids
+     * @returns {Promise} promise
```
reneolivo: should specify the resolve value. Can be `@returns {Promise<number[]>}` a promise that resolves to a list of award ids`.

ang/test/civiawards/dashboard/directives/more-filters-dashboard-action-button.controller.spec.js
```diff
@@ -0,0 +1,80 @@
...
+      });
+
+      it('opens a popup to select filters', () => {
+        expect(dialogService.open).toHaveBeenCalledWith('MoreFilters', '~/civiawards/dashboard/directives/more-filters-popup.html', jasmine.any(Object), {
+          autoOpen: false,
+          height: 'auto',
+          width: '350px',
+          title: 'More Filters'
+        });
+      });
```
reneolivo: I think we are missing tests for when we apply the filters.

ang/civiawards/dashboard/directives/dashboard-action-separator.js
deb1990: As discussed, removed the separator.

ang/civiawards/dashboard/directives/more-filters-dashboard-action-button.controller.js
deb1990: As discussed not changing it.

PR #47, opened 2020-01-06
RSE-657: Award buttons permissions

reneolivo:
## Overview

This PR only displays the "Create New Award" button, and the edit award cogs when the user has permission to create or edit awards only.

## Before & After

<table>
  <tr>
    <th></th>
    <th>No Permission</th>
    <th>With Permission</th>
  </tr>
  <tr>
    <th>Before</th>
    <td></td>
    <td></td>
  </tr>
  <tr>
    <th>After</th>
    <td>

</td>
    <td>

</td>
  </tr>
</table>

## Technical Details

This PR uses the permission introduced in https://github.com/compucorp/uk.co.compucorp.civiawards/pull/33

CiviCRM does not automatically add every permission to the `CRM.permissions` variable so we needed to manually expose it:

```php
  Civi::resources()
    ->addPermissions([
      'create/edit awards',
    ]);
```

The configuration files for adding the buttons use `CRM.checkPerm` to determine if they should or should not add the respective buttons.

### DIFF
--- a/ang/civiawards/shared/configs/case-type-buttons.config.js
+++ b/ang/civiawards/shared/configs/case-type-buttons.config.js
@@ -1,29 +1,35 @@
-(function (angular, _, url) {
+(function (angular, _, checkPermission, url) {
   var module = angular.module('civiawards');
   var AWARDS_CATEGORY_NAME = 'awards';
   var AWARD_CONFIG_URL = 'civicrm/a/#/awards/';
+  var EDIT_AWARD_PERMISSION = 'create/edit awards';
 
   module.config(function (CaseTypeProvider, CaseTypeCategoryProvider, DashboardCaseTypeButtonsProvider) {
     var awardCategory = CaseTypeCategoryProvider.findByName(AWARDS_CATEGORY_NAME);
     var awardCaseTypes = CaseTypeProvider.getByCategory(awardCategory.value);
+    var canEditAwards = checkPermission(EDIT_AWARD_PERMISSION);
 
-    addConfigurationButtonsToCaseTypes(awardCaseTypes);
+    (function () {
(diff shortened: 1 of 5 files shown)


ang/civiawards/shared/configs/case-type-buttons.config.js
```diff
@@ -1,29 +1,35 @@
...
     var awardCaseTypes = CaseTypeProvider.getByCategory(awardCategory.value);
+    var canEditAwards = checkPermission(EDIT_AWARD_PERMISSION);
 
-    addConfigurationButtonsToCaseTypes(awardCaseTypes);
+    (function () {
+      if (!canEditAwards) {
+        return;
+      }
+
+      _.forEach(awardCaseTypes, addConfigurationButtonsToCaseType);
```
deb1990: Why did we took the loop out of the function?

reneolivo: makes the function itself simpler. It also explains what happens to each award.

Imagine we need to filter by awards from last year first before adding the configuration button (a crazy example, bear with me). We can have  this on the `init` function:

```js
_.chain(awardCaseTypes)
  .filter(getLastYearAwards)
  .forEach(addConfigurationButtonsToCaseType);
```

This explains very well how we assign the configuration buttons.

It also allows us to use the function for a single object: `addConfigurationButtonsToCaseType(myAward)` or for multiple ones: `myAwards.forEach(addConfigurationButtonsToCaseType)`. With the previous implementation it was all or nothing.

reneolivo:
Closed in favour of https://github.com/compucorp/uk.co.compucorp.civiawards/pull/48 since this had the wrong branch name and the implementation changed.

PR #54, opened 2020-02-07
RSE-733: Add Application manager role for Awards

deb1990:
## Overview
Previously the People Tab under Case Details was broken for Award Type Cases. This PR fixes that by adding a `Application manager` role for all Awards.

## Before
2020-02-07 at 2 01 PM

## After
2020-02-07 at 2 00 PM

## Technical Details
Unlike Case Type of Non Award type, while creating Awards, the `CaseRoles` are not defined, which resulted in the People Tab being broken. To fix this, a role called `Application Manager` is created and the `manager` flag is set to true. Means people assigned to this role, will also be Award Application Manager.

```javascript
definition: {
  caseRoles: [{
    name: 'Application Manager',
    manager: 1
  }]
}
```

### DIFF
--- a/ang/civiawards/award-creation/directives/award.directive.js
+++ b/ang/civiawards/award-creation/directives/award.directive.js
@@ -202,7 +202,13 @@
         description: $scope.basicDetails.description,
         is_active: $scope.basicDetails.isEnabled,
         case_type_category: CaseTypeCategory.findByName('awards').value,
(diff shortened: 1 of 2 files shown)


tunbola:
@deb1990 , We might need to create the actual relationship `Application Manger` relationship while installing the extension.

PR #55, opened 2020-02-10
RSE-856: Add Access Review Custom Field Set Permission & Functionality

tunbola:
## Overview
Currently when the Award manager is not able to view or add review custom fields to an award except the Award manager has the `administer CiviCRM` and the `access all custom data` permissions. In the real scenario, the Award manager will not have access to these two permissions (at least the `administer CiviCRM` permission). This PR adds a new permission `access review custom field set` that allows the Award manager to view Award review custom fields only (i.e Custom fields belonging to a Custom Group that belongs to the reserved Applicant Review activity type).

## Before
Award manager cannot view not add review custom fields without having both the `administer CiviCRM` and the `access all custom data` permissions.

Configure Award - Latest Award  CiviAwards 2020-02-10 14-50-06

## After
Award manager can view and add review custom fields when assigned the new permission : `access review custom field set`

Configure Award - Latest Award  CiviAwards 2020-02-10 14-52-13

## Technical Details
- The `civiawards_civicrm_alterAPIPermissions` is used for this implementation. A new function `modifyCustomFieldApiPermission` is added to the `CRM_CiviAwards_Hook_alterAPIPermissions_Award` class ensures that the permission for the CustomField.get API is only modified if the Custom group for the custom fields only extend the Applicant Review activity type. 

```php
    if ($this->modifyCustomFieldApiPermission($entity, $action, $params)) {
      $permissions['custom_field']['get'] = [
        ['access review custom field set', 'access all custom data'],
      ];
    }
```

### DIFF
--- a/CRM/CiviAwards/Setup/CreateApplicantReviewActivityType.php
+++ b/CRM/CiviAwards/Setup/CreateApplicantReviewActivityType.php
@@ -23,7 +23,7 @@ class CRM_CiviAwards_Setup_CreateApplicantReviewActivityType {
     CRM_Core_BAO_OptionValue::ensureOptionValueExists([
       'option_group_id' => 'activity_category',
       'name' => self::APPLICANT_REVIEW,
-      'label' => self::APPLICANT_REVIEW,
+      'label' => ts(self::APPLICANT_REVIEW),
       'is_default' => 1,
       'is_active' => TRUE,
       'is_reserved' => TRUE,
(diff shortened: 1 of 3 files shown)


CRM/CiviAwards/Setup/CreateApplicantReviewActivityType.php
```diff
@@ -5,6 +5,8 @@
  */
 class CRM_CiviAwards_Setup_CreateApplicantReviewActivityType {
 
+  const APPLICANT_REVIEW = 'Applicant Review';
```
erawat: Don't we need localisation tag for this?

civiawards.php
```diff
@@ -171,6 +169,14 @@ function civiawards_civicrm_permission(&$permissions) {
     'CiviAwards: Create/Edit awards',
     ts('Allows a user to create or edit awards'),
   ];
+
+  $permissions['access review custom field set'] = [
+    'CiviAwards: Access review fields ',
```
erawat: Same as above, don't we need localisation tag for permission name?

tunbola: I'll add for this.

CRM/CiviAwards/Setup/CreateApplicantReviewActivityType.php
tunbola: Will add for where this is used in labels.

PR #60, opened 2020-02-20
RSE-842: Display source contact on review form

reneolivo:
## Overview
This PR allows to modify and view the source contact for a particular review.

### Creating a new review
Screen Shot 2020-02-20 at 11 40 00 AM

### Viewing and updating an existing review
Screen Shot 2020-02-20 at 11 40 31 AM
Screen Shot 2020-02-20 at 11 40 36 AM

### Creating a new review
pr-after-1

### Viewing and updating an existing review
pr-after-2

## Technical Details
We modified the review form class.

* Removed the `getActivityCaseId` function and replaced it with a `getActivity` function. The reason for this was to avoid making multiple calls just to get different fields of the same activity.
* For the default values we use the current logged in user when creating a new review, but the stored source contact when editing the review.

## Extra notes:

* We tried to replicate the experience from the regular activity form as much as possible hence the field is called "Reported By" when creating or editing the field, but "Added By" when viewing the field.

### DIFF
--- a/CRM/CiviAwards/Form/AwardReview.php
+++ b/CRM/CiviAwards/Form/AwardReview.php
@@ -73,10 +73,7 @@ class CRM_CiviAwards_Form_AwardReview extends CRM_Core_Form {
     $isAddAction = $this->_action & CRM_Core_Action::ADD;
 
     if ($this->activityId) {
-      $this->defaultValues = array_merge(
-        $this->defaultValues,
-        $this->getProfileFieldValues()
-      );
+      $this->defaultValues = $this->getProfileFieldValues();
     }
 
     $this->defaultValues['source_contact_id'] = $isAddAction
--- a/CRM/CiviAwards/Form/AwardReview.php
+++ b/CRM/CiviAwards/Form/AwardReview.php
@@ -63,11 +63,21 @@ class CRM_CiviAwards_Form_AwardReview extends CRM_Core_Form {
    * {@inheritDoc}
    */
   public function setDefaultValues() {
-    if (empty($this->defaultValues) && $this->activityId) {
-      $this->defaultValues = $this->getProfileFieldValues();
+    $this->defaultValues = [];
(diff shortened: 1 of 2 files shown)


CRM/CiviAwards/Form/AwardReview.php
```diff
@@ -63,11 +63,21 @@ class CRM_CiviAwards_Form_AwardReview extends CRM_Core_Form {
...
    */
   public function setDefaultValues() {
-    if (empty($this->defaultValues) && $this->activityId) {
-      $this->defaultValues = $this->getProfileFieldValues();
-
-      return $this->defaultValues;
+    $this->defaultValues = [];
+    $isAddAction = $this->_action & CRM_Core_Action::ADD;
+
+    if ($this->activityId) {
```
tunbola: Any reason why we can't have the original if condition `if (empty($this->defaultValues) && $this->activityId)`.
The reason is that the preprocess might get called a couple of time (e.g when submitting) and no need to set the default values again if it is already set.

reneolivo: I was not aware of this!

Since we need to set some values depending if the form is in editing mode and some others for create mode, I have made it so that the function is finished immediately to avoid setting the values again.

CRM/CiviAwards/Form/AwardReview.php
```diff
@@ -63,11 +63,27 @@ class CRM_CiviAwards_Form_AwardReview extends CRM_Core_Form {
...
-      return $this->defaultValues;
+    if ($hasDefaultValues) {
+      return;
+    }
+
+    $this->defaultValues = [];
+    $isAddAction = $this->_action & CRM_Core_Action::ADD;
+
+    if ($this->activityId) {
+      $this->defaultValues = array_merge(
```
tunbola: Do we really need a merge here. Why not just `$this->defaultValues = $this->getProfileFieldValues()` ?

PR #64, opened 2020-03-04
RSE-959: Move awards creation page to a new namespace

deb1990:
## Overview
Datepicker inside the Awards Creation/Edit page looked much bigger than normal. This PR fixes the same.

## Before
2020-03-04 at 3 56 PM

## After
2020-03-04 at 3 56 PM

## Technical Details
The datepicker was bigger than normal shoreditch. This is because the awards creation page was under `civicrm/a` namespace, so unnecessary css files were overriding styles.

So instead of `civicrm/a/#/awards/new`, the url is changed to `civicrm/award/a/#/awards/new`.

This way only shoreditch and awards css files are loaded.

1. Created `CRM/CiviAwards/Page/AwardAngular.php`, similar to `CRM/Civicase/Page/CaseAngular.php`.

2. Added a new path to use the above Page 
```xml
<item>
  <path>civicrm/award/a</path>
  <page_callback>CRM_CiviAwards_Page_AwardAngular</page_callback>
  <access_arguments>create/edit awards</access_arguments>
</item>
```


CRM/CiviAwards/Page/AwardAngular.php
```diff
@@ -0,0 +1,44 @@
...
+  public function run() {
+    $loader = new AngularLoader();
+    $loader->setPageName('civicrm/award/a');
+    $loader->setModules(['crmApp', 'civiawards']);
+    $loader->load();
+    \Civi::resources()->addSetting([
+      'crmApp' => [
+        'defaultRoute' => '/awards/new',
+      ],
+    ]);
```
tunbola: There should be a blank line before return. Maybe you forgot to run the syntax checker?

CRM/CiviAwards/Page/AwardAngular.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+use Civi\Angular\AngularLoader as AngularLoader;
+
+/**
+ * Class CRM_CiviAwards_Page_AwardAngular.
+ *
+ * Define an Angular base-page for CiviAwards.
+ */
+class CRM_CiviAwards_Page_AwardAngular extends \CRM_Core_Page {
```
tunbola: No need for a slash before `CRM_Core_Page`

PR #65, opened 2020-03-04
RSE-868: Fix Table Column width for review tab

deb1990:
## Overview
Inside the Review tab while creating and award, and also inside the Case Details section, the "Required" column and the Action column was taking more width than its content, this PR fixes the same.

## Before
2020-03-04 at 5 25 PM

2020-03-04 at 5 26 PM

## After
2020-03-04 at 5 26 PM

2020-03-04 at 5 25 PM

## Technical Details
Added static width in the markup.


jamienovick:
@deb1990  when we do the screenshots lets do them at 1280px as thats the target width and we can see any clipping issues... no need to change this but I can see very high res monitors are being used which aren't the way that users will be using it.

PR #66, opened 2020-03-04
RSE-652: Set API Permissions for AwardReviewPanel Entity

tunbola:
## Overview
An Award manager cannot add review panels currently because the default API permission `Administer Civicrm` is required. This PR fixes that and sets appropriate permissions for the AwardReviewPanel API's

## Before
- Award Manager cannot create a review panel.

## After
- Award Manager can now create review panels.
The API permissions for AwardReviewPanel Get/Create/Update is set to permissions required for creating an Award.

ReviewPanel

### DIFF
--- a/CRM/CiviAwards/Hook/AlterAPIPermissions/Award.php
+++ b/CRM/CiviAwards/Hook/AlterAPIPermissions/Award.php
@@ -37,6 +37,8 @@ class CRM_CiviAwards_Hook_AlterAPIPermissions_Award {
     ];
     $permissions['award_detail']['create'] = $permissions['award_detail']['update'] = $awardCreatePermission;
     $permissions['award_manager']['get'] = $awardCreatePermission;
+    $permissions['award_review_panel']['get'] = $awardCreatePermission;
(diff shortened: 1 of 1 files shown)


CRM/CiviAwards/Hook/AlterAPIPermissions/Award.php
```diff
@@ -37,6 +37,8 @@ public function run($entity, $action, array &$params, array &$permissions) {
     ];
     $permissions['award_detail']['create'] = $permissions['award_detail']['update'] = $awardCreatePermission;
     $permissions['award_manager']['get'] = $awardCreatePermission;
+    $permissions['award_review_panel']['get'] = $awardCreatePermission;
+    $permissions['award_review_panel']['create'] = $permissions['award_review_panel']['update'] = $awardCreatePermission;
```
reneolivo: it's better to have each permission assignation in its own line (like the `get` one above), so this is better:

```php
$permissions['award_review_panel']['get'] = $awardCreatePermission;
$permissions['award_review_panel']['create'] = $awardCreatePermission;
$permissions['award_review_panel']['update'] = $awardCreatePermission;
``

PR #68, opened 2020-03-05
RSE-678: Review Panel Group Selection UI + Functionality

deb1990:
## Overview
As part of this PR, the Review Panel Group Selection UI and functionality has been implemented.

## Screenshot
after

## Technical Details
1.`AwardReviewPanel` api now accepts array for the `contact_id` field. Because in the UI, more than one contact can be selected for Specific Relationships.

2. Files inside `ang/civiawards/award-creation/directives` has been reorganised. Files related to `review-fields` tab has been moved inside a folder with same name. Similar changes are done for files related to `review-panels`.

3. A new directive `contacts-group-dropdown` has been created to display the group selected dropdown. The code has been mostly copied from https://github.com/civicrm/civicrm-core/blob/master/ang/crmMailing/Recipients.js.

`crmMailingRecipients` is defined in core, but it is tightly coupled with Mailing related features, hence it was required to create a new directive instead of reusing the former.

2020-03-06 at 1 52 PM

## Comments
Unit tests for `contacts-group-dropdown` are not added yet, as because of review comments, we might need to change the unit tests, so this way we are trying to save some time, and write the unit tests part of RSE-658.


ang/civiawards/award-creation/directives/review-fields/review-fields-table.directive.js
```diff
@@ -175,7 +175,11 @@
      * Open the Popup to Select Review fields
      */
     function openReviewFieldSelectionPopup () {
-      dialogService.open('ReviewFields', '~/civiawards/award-creation/directives/review-field-selection.html', $scope, {
+      if (dialogService.dialogs.ReviewFields) {
```
reneolivo: is this here to avoid opening the same popup twice? But, why would it be opened twice?

scss/components/_award.scss
```diff
@@ -79,3 +85,23 @@
   min-height: 10px;
   min-width: 10px;
 }
+
+.civiawards__award__review-panel__to-label {
+  float: left;
+  line-height: 28px;
```
reneolivo: Why do we need this non standard line height here instead of using `$line-height` variable from shoreditch?

ang/civiawards/award-creation/directives/review-fields/review-fields-table.directive.js
deb1990: Yes. If the user clicks the Add Review Panel, while its already open, it will create the popup multiple times.

ang/civiawards/shared/directives/contacts-group-dropdown.directive.js
```diff
@@ -0,0 +1,193 @@
...
+        _.each(values.include, function (v) {
+          r.push('i' + v);
+        });
+
+        _.each(values.exclude, function (v) {
+          r.push('e' + v);
+        });
+
+        return r;
+      }
```
reneolivo: Several issues here:
- Can we remove these single variables and more descriptive ones?
- Why `'i' + v` ? can it be `'include_' + contact_id` etc ?

I think that can be refactored like this:

```js
function convertGroupsToValues (values) {
  return _.concat([],
    _.map(values.include, function (contact_id) { return 'include_' + contact_id; }),
    _.map(values. exclude, function (contact_id) { return 'exclude_' + contact_id; })
  );
}
```

ang/civiawards/shared/directives/contacts-group-dropdown.directive.js
```diff
@@ -0,0 +1,193 @@
...
+      (function init () {
+        ngModel.$render = refreshUI;
+
+        fetchGroupsData()
+          .then(prepareGroupsDataForSelect2)
+          .then(initSelect2)
+          .then(refreshUI);
+
+        scope.$watchCollection('recips.include', refreshUI);
+        scope.$watchCollection('recips.exclude', refreshUI);
```
reneolivo: is it possible to avoid using watch at all? Like adding an `ng-change` or similar to the view? Also, we don't need to watch for a specific inclusion/exclusion, it should trigger the refresh on any change.

ang/civiawards/shared/directives/contacts-group-dropdown.directive.js
```diff
@@ -0,0 +1,193 @@
...
+
+        elem.on('select2-selecting', selectEventHandler);
+
+        $(elem).on('select2-removing', removeEventHandler);
+      }
+
+      /**
+       * @param {Array} groupsData groups data
+       */
+      function prepareGroupsDataForSelect2 (groupsData) {
```
reneolivo: missing description for this function

reneolivo: Also, can we make it so it returns information instead of modifying a parameter?

ang/civiawards/shared/directives/contacts-group-dropdown.directive.js
```diff
@@ -0,0 +1,193 @@
...
+          return groupsData.values;
+        });
+      }
+
+      /**
+       * Select event handler for dropdown
+       *
+       * @param {object} e event object
+       */
+      function selectEventHandler (e) {
```
reneolivo: no single variable name.

ang/civiawards/shared/directives/contacts-group-dropdown.directive.js
```diff
@@ -0,0 +1,193 @@
...
+        $(elem).select2('close');
+        e.preventDefault();
+      }
+
+      /**
+       * Remove event handler for dropdown
+       *
+       * @param {object} e event object
+       */
+      function removeEventHandler (e) {
```
reneolivo: no single variable name.

ang/civiawards/shared/directives/contacts-group-dropdown.directive.js
```diff
@@ -0,0 +1,193 @@
...
+      }
+
+      /**
+       * Remove value from sent array
+       *
+       * @param {Array} array array from where value will be removed
+       * @param {number} value value to be removed
+       */
+      function arrayRemove (array, value) {
+        var idx = array.indexOf(value);
```
reneolivo: can we use `index` instead of `idx` ?

scss/components/_award.scss
deb1990: Fixed it, using `$crm-gap-small`.

PR #71, opened 2020-03-16
RSE-905: Case Category Custom Fields

tunbola:
## Overview
This PR does some cleanup in the civiawards extension due to the fact that there is a new approach in the way custom fields are handled for case type categories. See https://github.com/compucorp/uk.co.compucorp.civicase/pull/390.

## Before
- The award custom fields works based on the old way of implementing case category custom fields. 

## After
- The newly created `CaseCategoryCustomDataType` and `CaseCategoryCustomFieldExtends` classes are used to setup the Awards custom field support functionality


CRM/CiviAwards/Setup/ProcessAwardsCategoryForCustomGroupSupport.php
```diff
@@ -0,0 +1,24 @@
...
+use CRM_CiviAwards_Helper_CaseTypeCategory as CaseTypeCategoryHelper;
+use CRM_Civicase_Service_CaseCategoryCustomDataType as CaseCategoryCustomDataType;
+use CRM_Civicase_Service_CaseCategoryCustomFieldExtends as CaseCategoryCustomFieldExtends;
+
+/**
+ * Class CRM_CiviAwards_Setup_ProcessAwardsCategoryForCustomGroupSupport.
+ */
+class CRM_CiviAwards_Setup_ProcessAwardsCategoryForCustomGroupSupport {
+
+  const AWARDS_CATEGORY_CG_LABEL = 'Case (Awards)';
```
erawat: Do we need to localise Label?

tunbola: Not needed @erawat , Civi does that already in forms where the options are displayed. This value is only for inserting in DB

PR #72, opened 2020-03-17
RSE-672: Review Panel List UI + functionality

deb1990:
## Overview
As part of this PR, the Review Panel List, and also the Edit existing review panel has been developed.

## How it looks
after

## Technical Details
1. In `review-panels.directive.html` markup for Review Panel list has been added.
2. In `review-panels.directive.js` code to fetch the review panel list, and also to edit existing review panels has been added.
    * ` $scope.reviewPanel` has been renamed to ` $scope.currentReviewPanel` to make it more readable.


ang/civiawards/award-creation/directives/review-panels/review-panels.directive.js
```diff
@@ -13,28 +13,213 @@
...
+        existingReviewPanels: fetchExistingReviewPanels($scope.awardId)
+      }).then(function (fetchedData) {
+        if (fetchedData.existingReviewPanels && fetchedData.existingReviewPanels.length === 0) {
+          return fetchedData;
+        }
+
+        return fetchContactsFromPanelsAndIndexById(fetchedData.existingReviewPanels)
+          .then(function () {
+            return fetchedData;
+          });
```
reneolivo: there are two issues with `fetchContactsFromPanelsAndIndexById`:

1 - it does not return anything. It's doing something hidden and we need to peek at the function to understand it.
2 - it's doing 2 things at the same time: it fetches data, and applies some indexing (even if it's minimal)

I believe the best action would be to append the response of `fetchContactsFromPanels` to `fetchedData`:

```js
.then(function (fetchedData) {
  var shouldFetchContactsFromPanels = fetchedData.existingReviewPanels && 
    fetchedData.existingReviewPanels.length === 0;

  if (!shouldFetchContactsFromPanels) {
    return fetchedData;
  }

  return fetchContactsFromPanels(fetchedData.existingReviewPanels)
    .then(function (contactsFromPanels) {
      return _.extend({}, fetchedData, {
        contactsFromPanels: contactsFromPanels
      });
    });
})
.then(function (fetchedData) {
  contactsIndexed = _.indexBy(fetchedData.contactsFromPanels, 'id');
  groupsIndexed = _.indexBy(fetchedData.groups, 'id');
  // ...
});
```

As a note, if you want to make a specialized function you could have this instead:

```js
fetchContactsFromPanels(someData)
  .then(storeContactsIndexedById);
```

PR #73, opened 2020-03-17
RSE-671: Delete Review Panels + Enable/Disable from list

deb1990:
## Overview
As part of this PR, the Review Panel Delete and Enabled/Disable feature has been developed.

## How it looks
after

## Technical Details
1. In `review-panels.directive.html` added Enable/Disable and Delete Dropdowns.
2. In `review-panels.directive.js` functions has been added to handle the above features.


ang/civiawards/award-creation/directives/review-panels/review-panels.directive.js
```diff
@@ -305,16 +410,21 @@
...
-
-        return {
-          is_a_to_b: isAToB ? '1' : '0',
-          relationship_type_id: relationshipTypeId,
-          contact_id: getSelect2Value(relation.contacts)
-        };
-      });
+      return _.chain($scope.currentReviewPanel.relationships)
+        .filter(function (relation) {
+          return relation.type.length !== 0;
```
reneolivo: why was this filter introduced? How come a relationship type can be empty?

ang/test/civiawards/award-creation/directives/review-panels/review-panels.directive.spec.js
```diff
@@ -322,6 +323,161 @@
...
+        });
+      });
+
+      describe('when yes is pressed on confirmation prompt', () => {
+        beforeEach(() => {
+          crmConfirmCallbackFn();
+          $scope.$digest();
+        });
+
+        it('deletes the review panel if yes is pressed', () => {
```
reneolivo: `it('deletes the review panel', () => {`

No need to add the condition in the expectation.

ang/test/civiawards/award-creation/directives/review-panels/review-panels.directive.spec.js
```diff
@@ -322,6 +323,161 @@
...
+        });
+      });
+
+      describe('when yes is pressed on confirmation prompt', () => {
+        beforeEach(() => {
+          crmConfirmCallbackFn();
+          $scope.$digest();
+        });
+
+        it('enables the review panel if yes is pressed', () => {
```
reneolivo: remove "if yes is pressed"

ang/test/civiawards/award-creation/directives/review-panels/review-panels.directive.spec.js
```diff
@@ -322,6 +323,161 @@
...
+        });
+      });
+
+      describe('when yes is pressed on confirmation prompt', () => {
+        beforeEach(() => {
+          crmConfirmCallbackFn();
+          $scope.$digest();
+        });
+
+        it('disables the review panel if yes is pressed', () => {
```
reneolivo: remove "if yes is pressed"

scss/components/_award.scss
```diff
@@ -1,3 +1,9 @@
+.___outside-namespace {
+  .civiawards__award__review-panel-form__delete {
+    float: left !important;
```
reneolivo: why is the `!important` needed? Can it be avoided by adding more specificity ?

ang/civiawards/award-creation/directives/review-panels/review-panels.directive.js
deb1990: Because we always have an empty relationship added so that user can enter data, but is should not be saved if its empty. So we remove it before saving.

2020-03-19 at 11 00 AM

PR #75, opened 2020-03-24
RSE-1033: Add disabled case type filter

reneolivo:
## Overview
This PR allows filtering applications belonging to disabled awards.

**Note:** this PR is linked to https://github.com/compucorp/uk.co.compucorp.civicase/pull/402

## Before
Screen Shot 2020-03-23 at 10 26 35 PM
the filter does not exist.

## After
pr-after

## Technical Details

First, we needed to modify the `AwardManager.xml` file to be able to create a real relationship to case types. Previously we were storing the case type id value here, but could not do sub queries such as `crmApi('AwardManager', 'get', { 'case_type_id.is_active': '1' })`. After modifying the file we ran `civix generate:entity-boilerplate` and it automatically updated the DAO and SQL files attached to this PR.

For the checkbox element we use `civicase__checkbox`.

The "more filters" action button will pass either `'1'` or `'0'` if this new filter is selected or not. The value is used for both: filtering the award manager case types, and the case types shown on the overview section.


ang/civiawards/dashboard/directives/more-filters-dashboard-action-button.controller.js
```diff
@@ -91,7 +91,10 @@
         .then(processAwardTypeFilters)
         .then(function (awardTypeIds) {
           var param = {
-            case_type_id: awardTypeIds.length > 0 ? { IN: awardTypeIds } : ''
+            case_type_id: awardTypeIds.length > 0 ? { IN: awardTypeIds } : '',
```
deb1990: If we are already having the same check inside `processMyAwardsFilter`, do we need this filter anymore? Because `processMyAwardsFilter` will return only those Case Types which are Active(or not active), and we use those Case Type IDS to fetch further items in dashboard, so not sure why this is needed.

ang/civiawards/dashboard/directives/more-filters-popup.html
```diff
@@ -23,6 +23,21 @@
 
       <label>Award End Date</label>
       <civicase-ui-date-range date-range="model.selectedFilters.end_date"></civicase-ui-date-range>
+      <br />
```
deb1990: Do we need `<br>`? The rest of the elements have a line break without using `<br>`

ang/civiawards/dashboard/directives/more-filters-popup.html
```diff
@@ -23,6 +23,21 @@
...
+      <label
+        class="civicase__case-filter-form-elements"
+        ng-click="model.selectedFilters.showDisabledAwards = !model.selectedFilters.showDisabledAwards"
+      >
+        <span class="civicase__checkbox" role="checkbox" aria-checked="true" tabindex="0">
+          <i
+            ng-if="model.selectedFilters.showDisabledAwards"
+            class="civicase__checkbox--checked material-icons ng-scope"
```
deb1990: `ng-scope` :O

sql/auto_install.sql
```diff
@@ -28,17 +81,23 @@ SET FOREIGN_KEY_CHECKS=1;
 -- *
 -- *******************************************************/
 CREATE TABLE `civicrm_civiawards_award_detail` (
```
tunbola: @reneolivo , can you clean up this file a bit so it looks better. The Civicrm comments above can be removed, we don't need it. The SQL queries in this file also does not have any formatting and are scattered all over the place, can you format it similar to how it was before the change?  Thanks.

sql/auto_uninstall.sql
```diff
@@ -1,3 +1,30 @@
...
+-- | License and the CiviCRM Licensing Exception along                  |
+-- | with this program; if not, contact CiviCRM LLC                     |
+-- | at info[AT]civicrm[DOT]org. If you have questions about the        |
+-- | GNU Affero General Public License or the licensing of CiviCRM,     |
+-- | see the CiviCRM license FAQ at http://civicrm.org/licensing        |
+-- +--------------------------------------------------------------------+
+--
+-- Generated from drop.tpl
+-- DO NOT EDIT.  Generated by CRM_Core_CodeGen
+--
```
tunbola: We can remove the comments above too. We don't need it.

ang/civiawards/dashboard/directives/more-filters-popup.html
reneolivo: Screen Shot 2020-03-24 at 7 50 09 AM
@deb1990 other inputs look nice without the `<br />` because they have their labels above them, and this creates some spacing. The checkbox has the label next to it so there's no spacing and it looks like it's part of the end date fields.

ang/civiawards/dashboard/directives/more-filters-popup.html
reneolivo: 🏃 💨

ang/civiawards/dashboard/directives/more-filters-dashboard-action-button.controller.js
reneolivo: @deb1990 this is needed because even if we send specific case type IDs, the dashboard overview will directive will request active case types:
https://github.com/compucorp/uk.co.compucorp.civicase/pull/402/files#diff-6576be8656588616fa32fd3f5e72c5b9R171

So this basically is forcing the case overview to display disabled ones.

Also, we need the case overview to force enabled case types because that's the default behaviour, otherwise it would show all cases.

PR #77, opened 2020-03-24
RSE-944: Fix API param value when case type id is null

swastikpareek:
## Overview
This PR is a supportive PR for the changes introduced in https://github.com/compucorp/uk.co.compucorp.civicase/pull/404. This basically contains the related FE changes done on awards side.

## Before
RSE-944-before

## After
RSE-944-after

## Technical Details
The `getStats` API is improved to support `IS NULL` operator for `case_type_id` parameter. See the technical details of https://github.com/compucorp/uk.co.compucorp.civicase/pull/404.
Related FE changes are being incorporated here

### DIFF
--- a/ang/test/civiawards/dashboard/directives/more-filters-dashboard-action-button.controller.spec.js
+++ b/ang/test/civiawards/dashboard/directives/more-filters-dashboard-action-button.controller.spec.js
@@ -3,7 +3,7 @@
 (function (_) {
   describe('More Filters Dashboard Action Button', () => {
     let $q, $location, $scope, $controller, $rootScope, dialogService,
-      crmApiMock, crmApi;
+      crmApiMock, originalCrmAPIMock, crmApi;
 
     beforeEach(module('civiawards', function ($provide) {
       crmApiMock = jasmine.createSpy();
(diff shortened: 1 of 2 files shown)


deb1990:
@swastikpareek Labels and Unit tests are missing. Please fix.

ang/test/civiawards/dashboard/directives/more-filters-dashboard-action-button.controller.spec.js
```diff
@@ -195,6 +195,32 @@
           expect($scope.isNotificationVisible()).toBe(true);
         });
       });
+
+      describe('when filters response yields no awards types ids', () => {
+        beforeEach(() => {
+          originalCrmAPIMock = crmApiMock;
```
deb1990: In this case storing and resetting `crmApiMock` is not necessary, because in Line 9, it will be reset for every unit test anyways. Resetting is only necessary when you are updating a global variable(like CRM), which not an angular component. As angular components are set in beforeEach section, they do not need special attention.

ang/test/civiawards/dashboard/directives/more-filters-dashboard-action-button.controller.spec.js
```diff
@@ -195,6 +195,32 @@
...
+            values: []
+          }));
+        });
+
+        beforeEach(() => {
+          dialogModel.applyFilterAndCloseDialog();
+          $rootScope.$digest();
+        });
+
+        afterEach(() => {
```
deb1990: As per above comment, this is not necessary anymore.

ang/test/civiawards/dashboard/directives/more-filters-dashboard-action-button.controller.spec.js
```diff
@@ -195,6 +195,32 @@
...
+          crmApiMock = originalCrmAPIMock;
+        });
+
+        it('should set correct param value for case_type_id param', () => {
```
deb1990: Lets not use statement like `set correct param value`. Instead explain what that param value is.
Also lets not use technical statements like `case_type_id`. Instead the unit test statement should be a valid natural UI state(if possible), so for example
`hides all case types, cases and activities from dashboard`.

So the whole Statement including the `describe` would be

`when filters response yields no awards types ids, it hides all case types, cases and activities from dashboard`

PR #79, opened 2020-03-25
RSE-654: Fix error when no Relationships are present in Review Panels

deb1990:
## Overview
When there are no defined Specific Relationship in all the Review panels of an award, the Review Panels cannot be seen. This PR fixed that issue.

## Before
2020-03-25 at 3 27 PM

2020-03-25 at 4 08 PM

## After
2020-03-25 at 3 28 PM

The `Contact.get` api call is not happening anymore.

## Technical Details
In `review-panels.directive.js`, after fetching the Review Panels, we fetch  the list of Contact names from the list of Relationships, but when the relationships are empty, we used to send an empty array to the `Contact.get` api call.
```javascript
return crmApi('Contact', 'get', {
	sequential: 1,
	return: ['display_name'],
	id: { IN: contactIdsToFetch }, // this used to be an empty array
	options: { limit: 0 }
})
.then(function (contactsData) {
	return contactsData.values;
});
```

So we were getting an `invalid criteria for IN` error.

To fix this, now we check if the list of contacts are empty, and if yes, we directly resolve the promise
```javascript
if (contactIdsToFetch.length === 0) {
    return $q.resolve([]);
}
```

### DIFF
--- a/ang/test/civiawards/award-creation/directives/review-panels/review-panels.directive.spec.js
+++ b/ang/test/civiawards/award-creation/directives/review-panels/review-panels.directive.spec.js
@@ -2,14 +2,8 @@
 (function (_) {
   describe('civiawardReviewPanels', () => {
     let $rootScope, $controller, $scope, crmApi, $q, crmStatus, ts, ContactsData,
-      dialogService, RelationshipTypeData, GroupData, ReviewPanelsMockData;
-    const entityActionHandlers = {
-      'AwardReviewPanel.create': _.noop,
-      'RelationshipType.get': relationshipTypeGetHandler,
-      'AwardReviewPanel.get': awardReviewPanelGetHandler,
-      'Contact.get': contactGetHandler,
-      'Group.get': groupGetHandler
-    };
+      dialogService, RelationshipTypeData, GroupData, ReviewPanelsMockData,
+      entityActionHandlers;
 
     beforeEach(module('civiawards.templates', 'civiawards', 'crmUtil', 'civicase.data', 'civiawards.data', function ($provide) {
       $provide.value('crmApi', getCrmApiMock());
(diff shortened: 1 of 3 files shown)


swastikpareek:
@deb1990 : I tried running angular unit test and 33 of them failed.
Screenshot 2020-03-25 at 3 56 06 PM
Could you please take a look ?

ang/test/civiawards/award-creation/directives/review-panels/review-panels.directive.spec.js
```diff
@@ -271,58 +266,138 @@
...
+          expect(crmApi).toHaveBeenCalledWith('AwardReviewPanel', 'get', {
+            sequential: 1,
+            case_type_id: '10',
+            options: { limit: 0 }
+          });
+        });
+
+        it('displays all existing review panels for the award', () => {
+          expect($scope.existingReviewPanels).toEqual([{
+            id: '46',
```
swastikpareek: Can we save this object into a new variable and store it in `mocks/data` folder ?

ang/test/civiawards/award-creation/directives/review-panels/review-panels.directive.spec.js
```diff
@@ -271,58 +266,138 @@
...
+            sequential: 1,
+            case_type_id: '10',
+            options: { limit: 0 }
+          });
+        });
+
+        it('displays all existing review panels for the award', () => {
+          expect($scope.existingReviewPanels).toEqual([{
+            id: '47',
+            title: 'New Panel 2',
```
swastikpareek: Same like the comment above. Do you think we should move it to `mocks/data` folder ?

deb1990: Moved to a new variable, as discussed.

ang/test/civiawards/award-creation/directives/review-panels/review-panels.directive.spec.js
deb1990: Moved to a new variable, as discussed.

PR #80, opened 2020-03-25
RSE-956: Define award activity types

reneolivo:
## Overview
This PR adds default activity types to awards so users can use them to create activities related to applications.

**Note:** this PR is linked to https://github.com/compucorp/uk.co.compucorp.civicase/pull/408

## Before
Screen Shot 2020-03-25 at 4 43 16 PM
The menu does not exist

## After
pr-after

## Technical Details

* We modified the `awards.directive.js` file so it always send the Email, Follow up, Meeting, and Phone Call activities when saving the award (case type) definition. These activities are manually selected because that's how core defines them instead of selecting the communication activity grouping, which includes extra activities.
* The `review-form.service.js` service was refactored so it allows for options overriding so we can display different kind of forms depending on the action (ie: add, update, and view forms)
* `civiawards.php` implements `hook_civicrm_alterMenu` so we can pass the `civicase_reload` argument to it. Without the form is not displayed at all, and we need this parameter to trigger a case refresh after the review is saved.

### DIFF
--- a/ang/civiawards/award-creation/directives/award.directive.js
+++ b/ang/civiawards/award-creation/directives/award.directive.js
@@ -16,14 +16,12 @@
   module.controller('CiviAwardCreateEditAwardController', function (
     $location, $q, $scope, $window, CaseTypeCategory, crmApi, crmStatus, getSelect2Value) {
     var ts = CRM.ts('civicase');
-    var DEFAULT_COMMUNICATION_ACTIVITY_TYPES = [
+    var DEFAULT_ACTIVITY_TYPES = [
+      { name: 'Applicant Review' },
       { name: 'Email' },
       { name: 'Follow up' },
       { name: 'Meeting' },
       { name: 'Phone Call' }
-    ];
-    var DEFAULT_REVIEW_ACTIVITY_TYPES = [
-      { name: 'Applicant Review' }
     ];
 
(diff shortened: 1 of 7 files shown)


ang/civiawards/award-creation/directives/award.directive.js
```diff
@@ -16,6 +16,15 @@
   module.controller('CiviAwardCreateEditAwardController', function (
     $location, $q, $scope, $window, CaseTypeCategory, crmApi, crmStatus, getSelect2Value) {
     var ts = CRM.ts('civicase');
+    var DEFAULT_COMMUNICATION_ACTIVITY_TYPES = [
```
deb1990: Is it worth it having two separate arrays? how about having a single one called `DEFAULT_ACTIVITY_TYPES`?

ang/test/civiawards/award-creation/directives/award.directive.spec.js
```diff
@@ -257,16 +271,30 @@
...
                 manager: 1
               }],
               statuses: ['Open']
-            },
+            }),
             id: '10'
           });
         });
+
+        it('keeps the default activity types', () => {
```
deb1990: How about `creates the award with default activity types`?

PR #84, opened 2020-04-01
RSE-1052: Hide delete button when creating new panel

deb1990:
## Overview
When creating a new Review Panel, the delete button is visible inside the Review Panel Popup, which is wrong, as it should only appear when editing an existing review panel. This PR fixes this bug.

## Before
**Create Mode**
2020-04-01 at 12 58 PM

**Edit Mode**
2020-04-01 at 12 19 PM

## After
**Create Mode**
2020-04-01 at 12 58 PM

**Edit Mode**
2020-04-01 at 12 18 PM

## Technical Details
1. Inside `review-panels/review-panels.directive.js`, added a new function `prepareButtonsForReviewPanelPopup`, which adds the Delete button only if the Award Review panel id is present(which means an existing review panel).
```javascript

if ($scope.currentReviewPanel.id) {
  buttons.push({
    text: ts('Delete'),
    icons: { primary: 'fa-times' },
    class: 'civiawards__award__review-panel-form__delete',
    click: function () {
      $scope.$apply(function () {
        handleDeleteReviewPanel($scope.currentReviewPanel);
      });
    }
  });
}
```

### DIFF
--- a/ang/test/civiawards/award-creation/directives/review-panels/review-panels.directive.spec.js
+++ b/ang/test/civiawards/award-creation/directives/review-panels/review-panels.directive.spec.js
@@ -197,7 +197,7 @@
 
       describe('delete button', () => {
         describe('when creating new review panel', () => {
-          it('hides the delete button inside the popup', () => {
+          it('displays only the save button inside the popup', () => {
             expect(dialogService.open).toHaveBeenCalledWith('ReviewPanels',
               '~/civiawards/award-creation/directives/review-panels/review-panel-popup.html',
               $scope,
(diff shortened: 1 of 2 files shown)


ang/test/civiawards/award-creation/directives/review-panels/review-panels.directive.spec.js
```diff
@@ -194,6 +194,54 @@
           }, jasmine.any(Object));
         });
       });
+
+      describe('delete button', () => {
+        describe('when creating new review panel', () => {
+          it('hides the delete button inside the popup', () => {
```
swastikpareek, line 200: The it says that it checks if the button is hidden, but logic checks if save button is present. 
I suggest we change the `it` statement to say 
`it('has only the save action button inside popup')`
and inside another it we should have `has both save and delete action button inside the popup`.

PR #98, opened 2020-06-08
RSE-1132: Award Review Form Changes

tunbola:
## Overview
We need to be able to use the Award Review form from within SSP (Currently only accessible within Civicrm) so that a review panel member can add review for an award.
Below are the changes we need to make to the awards review form, both general changes and SSP specific changes:

**SSP Specific changes**
- The review form for SSP must not be themed with shoreditch styles
- A user from SSP can not view the review another panel member left on an Award, i.e user should able to view only own reviews

**General Changes**
- Once a user has reviewed an Award, the review cannot be edited again.
- A user can only leave a single review for an Award.

## Before
These changes were not made.

## After
The changes described in the overview section was implemented

| Award Review (Civicrm)  | Award Review (SSP) |
| ------------- | ------------- |
| View Review - Benefit Specialist Benefit - Award Type  CaseLatest 2020-06-08 10-23-55| View Review - Benefit Specialist Benefit - Award Type  CaseLatest 2020-06-08 10-25-04|
| Add Review  CaseLatest 2020-06-08 10-27-05|Add Review  CaseLatest 2020-06-08 10-30-29|
| Warning  CaseLatest 2020-06-08 10-45-39|Warning  CaseLatest 2020-06-08 10-29-46|
| Can view other reviews|Warning  CaseLatest 2020-06-08 10-47-23|

## Technical Details
To achieve what was described in the overview:

- A new menu path was created that points to the review form class, this path uses a permission that a Drupal SSP user will have access to and also has the `is_public` property to be true so that Shoreditch styles are not loaded for this path.
- When the form is accessed via SSP, the user does not have access to Civicrm API so fields like the `Added by` field that uses the entity API to fetch contacts was disabled for the SSP user, also links to the Civicrm contact page for the Added by contact was also removed as the SSP user will not have civicrm contact access.
- The Edit action was disabled in the Award review form, so only the Add and View actions are active.

## Comments
A PHP lint error resulting from a new rule in the Drupal Coder rule set was excluded. Please see this [PR](https://github.com/compucorp/uk.co.compucorp.civicase/pull/483) for more details

### DIFF
--- a/CRM/CiviAwards/Form/AwardReview.php
+++ b/CRM/CiviAwards/Form/AwardReview.php
@@ -11,6 +11,16 @@ use CRM_CiviAwards_BAO_AwardDetail as AwardDetail;
 class CRM_CiviAwards_Form_AwardReview extends CRM_Core_Form {
 
   /**
+   * URL for accessing review form from SSP.
+   */
+  const SSP_REVIEW_URL = 'civicrm/award-review-ssp';
+
+  /**
+   * URL for accessing review form from civicrm.
+   */
+  const CIVICRM_REVIEW_URL = 'civicrm/awardreview';
+
+  /**
    * Case ID.
    *
    * @var int
@@ -198,7 +208,7 @@ class CRM_CiviAwards_Form_AwardReview extends CRM_Core_Form {
       CRM_Core_Session::setStatus(ts('An error occurred'), 'Error', 'error');
     }
 
-    $awardPath = $this->isReviewFromSsp() ? 'civicrm/award-review-ssp' : 'civicrm/awardreview';
+    $awardPath = $this->isReviewFromSsp() ? self::SSP_REVIEW_URL : self::CIVICRM_REVIEW_URL;
     $url = CRM_Utils_System::url($awardPath, [
       'action' => 'view',
       'id' => $activityId,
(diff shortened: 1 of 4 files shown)


CRM/CiviAwards/Form/AwardReview.php
```diff
@@ -10,6 +10,16 @@
  */
 class CRM_CiviAwards_Form_AwardReview extends CRM_Core_Form {
 
+  /**
+   * URL for accessing review form from SSP.
+   */
+  const SSP_REVIEW_URL = 'civicrm/award-review-ssp';
```
hitesh-jain, line 16: @tunbola can we change this to "civicrm/ssp/awardreview" pls

xml/Menu/civiawards.xml
```diff
@@ -3,9 +3,16 @@
...
     <page_callback>CRM_CiviAwards_Form_AwardReview</page_callback>
-    <title>AwardReview</title>
+    <title>Award Review</title>
     <access_arguments>access CiviCRM</access_arguments>
   </item>
+  <item>
+    <path>civicrm/award-review-ssp</path>
+    <page_callback>CRM_CiviAwards_Form_AwardReview</page_callback>
+    <title>Award Review</title>
+    <access_arguments>access applicant portal</access_arguments>
```
hitesh-jain, line 13: @tunbola we are using "CiviAwards: Access awards panel portal" in SSP Drupal for Applications review listing and individual pages so I think we should use same here ?

PR #100, opened 2020-06-11
RSE-1115: HTML format and class tweaks to achieve design layout

swastikpareek:
## Overview
As a part of this PR basic bootstrap classes and HTML is implemented in order to match it closely with the designs. 
Related FE PR [here](https://github.com/compucorp/ssp_bootstrap/pull/10)

## Before
Screenshot 2020-06-11 at 6 31 37 PM

#### Form
Screenshot 2020-06-11 at 6 31 05 PM

#### Form Submission
Screenshot 2020-06-16 at 12 40 51 PM

##### Civicrm
Screenshot 2020-06-16 at 12 44 12 PM

#### Error
Screenshot 2020-06-11 at 6 36 52 PM

## Technical Details
* Custom civicrm classes are removed and bootstrap classes are used to achieve out of the box styles.
* Also, replaced core civicrm action buttons HTML with bootstrap action buttons in order to improve UI.
* The UI mentioned in the after screenshot can be achieved by some extra css overrides done for civicrm in `ssp_bootstrap` [theme here](https://github.com/compucorp/ssp_bootstrap/pull/10).

## Comments
This HTML contains dummy data for form prefix field. This is done in order to assist BE with development. So, they just need to extract the value from the DB and update it in the HTML template.

## Update
In [this commit](https://github.com/compucorp/uk.co.compucorp.civiawards/pull/100/commits/8fcf32187841347972a175b2e4d650642857aa88), the placeholder HTML for awards name and application tags are updated with actual value. While implementing this it has been found out that `caseTags` variable only holds the name of the application tag and not the color. To fix this, `$caseTags` variable processing is updated and datatype has been changed from `String` to `Array` which contains `name` and `color` as key. 
 And to fix its implication, this function which was consuming it's string value (`getPageTitle`) is updated to work with `array` now as this array is processed and `name` key is extracted to retain its original value.

### DIFF
--- a/CRM/CiviAwards/Form/AwardReview.php
+++ b/CRM/CiviAwards/Form/AwardReview.php
@@ -248,6 +248,7 @@ class CRM_CiviAwards_Form_AwardReview extends CRM_Core_Form {
 
     $this->assign('caseContactDisplayName', $this->getCaseContactDisplayName());
     $this->assign('caseTypeName', $this->caseTypeName);
+    $this->assign('caseTags', $this->caseTags);
     $this->assign('isViewAction', $isViewAction);
     $this->assign('isReviewFromSsp', $this->isReviewFromSsp());
 
--- a/templates/CRM/CiviAwards/Form/AwardReview.tpl
+++ b/templates/CRM/CiviAwards/Form/AwardReview.tpl
@@ -1,55 +1,55 @@
-<div id="bootstrap-theme">
-  <div class="civiawards__review-activity crm-form-block">
-    <div class="panel panel-default">
-      <div class="panel-body">
-        <div class="form-group row">
-          {if $errorMessage}
-            <p class="alert alert-warning">
-              {$errorMessage}
-            </p>
-          {/if}
-          {if $isViewAction}
-            <div class="col-sm-5">
-              <label for="source_contact_id">
-                {ts}Added By{/ts}
-              </label>
-            </div>
-            <div class="col-sm-7">
-                {if !$isReviewFromSsp }
-              <a
-                class="view-contact no-popup"
-                href=
-                "{crmURL p="civicrm/contact/view" q="reset=1&cid=`$sourceContactId`"}"
-                title="{ts}View Contact{/ts}"
-              >
-                {$sourceContactName}
-              </a>
-              {else}
-                {$sourceContactName}
-              {/if}
-            </div>
-          {/if}
-          {if !$isViewAction}
-            <div class="col-sm-5">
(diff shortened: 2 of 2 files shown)


jamienovick:
@swastikpareek out of interest - how will this affect the equivalent screen when a CiviCRM administrator accesses it and views it via the Civicase module? Is that also bootstrap? Thanks

swastikpareek:
> @swastikpareek out of interest - how will this affect the equivalent screen when a CiviCRM administrator accesses it and views it via the Civicase module? Is that also bootstrap? Thanks

I need to check that.

swastikpareek:
> > @swastikpareek out of interest - how will this affect the equivalent screen when a CiviCRM administrator accesses it and views it via the Civicase module? Is that also bootstrap? Thanks
> 
> I need to check that.

Just checked @jamienovick : And it looks same for civicrm user, civicrm admin, civicase user, and awards manager user roles.

jamienovick:
@swastikpareek I'm not sure I understand

When it loads on the CiviCRM side the desired outcome would be to load "Shoreditch" theme (or to look like normal CiviCRM screens and not the SSP theme? Could you post a screenshot?

@fran-compucorp @shitijg @marcelocompucorp was this considered in the spec?

Thanks

swastikpareek:
> When it loads on the CiviCRM side the desired outcome would be to load "Shoreditch" theme (or
How can I do that. When I open this url it opens in Drupal.

tunbola:
@swastikpareek , I didn't know you will need to make changes to this template directly, i thought the styles will be applied somewhere in SSP. This same form is also used by the Civicrm administrator to add/view reviews. 

If it is necessary that you have to modify this template, it might be a good idea at this point to add a separate template for the SSP and keep the original version for civicrm. I am just concerned about maintainability of updating twom templates whenever there is a change.

Is there no way to style this form by having some external style sheet since it is already loaded without shoredicth?

tunbola:
Argh. Sorry. I mistakenly closed this @swastikpareek. :(. Is there a way to re-open?

jamienovick:
@tunbola @swastikpareek  - given that both forms are bootstrap is there a way to have them both use the same classes but have different styling applied depending on some sort of routing?

swastikpareek:
> @tunbola @swastikpareek - given that both forms are bootstrap is there a way to have them both use the same classes but have different styling applied depending on some sort of routing?

yes, we can have contextual templating. Which spits HTML differently for SSP and civicrm. I will be updating the PR soon.

PR #101, opened 2020-06-11
RSE-841: Display tag colours on award review form header

ivan-compucorp:
## Overview
There is **Award review** form which displays case tags in the title, but they are being displayed as normal text. This PR makes them displayed like colored badges.
*View review* and *Edit review* form popups are affected.

## Before
image

## After
image

### Back-end
The `CRM_CiviAwards_Form_AwardReview` class was updated to return colored tags:
1. As now we need to store some more data for case tags (not only name) we renamed `getCaseTags()` to `loadCaseTags()`.
2. The `getCaseTags()` method now returns list of tags in desired format: `plain` (plain text - as it previously was) or `badge` (html - colored badges).
3. As case tag data contain background color, but does not contain text color the `getTextColor()` method was added to return suitable text color for given background color.

### Front-end
By default html in popup title looks like plain text:
image

So to workaround this (to make it possible to display colored tags in the title) we updated the `reviews-case-tab-content.directive.js`:
1. The `popupTitleDecodeEntities()` with helper `htmlEntitiesDecode()` function are added.
2. When popup is loaded (`loadForm().on('crmFormLoad')` event is used) the `popupTitleDecodeEntities()` is called now to decode html and display colored tags. See `handleViewReviewActivity()` and `handleEditReviewActivity()`.

### DIFF
--- a/CRM/CiviAwards/Form/AwardReview.php
+++ b/CRM/CiviAwards/Form/AwardReview.php
@@ -386,9 +386,9 @@ class CRM_CiviAwards_Form_AwardReview extends CRM_Core_Form {
     $caseTags = [];
     foreach ($result['values'] as $caseTag) {
       $caseTag = [
-        'id'               => $caseTag['api.Tag.getsingle']['id'],
-        'parent_id'        => $caseTag['api.Tag.getsingle']['parent_id'],
-        'name'             => $caseTag['api.Tag.getsingle']['name'],
+        'id' => $caseTag['api.Tag.getsingle']['id'],
+        'parent_id' => $caseTag['api.Tag.getsingle']['parent_id'],
+        'name' => $caseTag['api.Tag.getsingle']['name'],
         'background_color' => !empty($caseTag['api.Tag.getsingle']['color']) ? $caseTag['api.Tag.getsingle']['color'] : '#ffffff',
       ];
       $caseTag['color'] = $this->getTextColor($caseTag['background_color']);
@@ -401,28 +401,17 @@ class CRM_CiviAwards_Form_AwardReview extends CRM_Core_Form {
   /**
    * Get Case tags list.
    *
-   * @param string $format
-   *   (Optional) Output format, defaults to plain text.
-   *
    * @return string
-   *   Case tags list.
+   *   HTML code of Case tags as badge list.
    */
-  private function getCaseTags($format = 'plain') {
-    switch ($format) {
-      case 'badge':
-        $res = [];
-        foreach ($this->caseTags as $caseTag) {
-          $res[] = '<span class="crm-tag-item" style="background-color: ' . $caseTag['background_color'] . '; color: ' . $caseTag['color'] . '">' . $caseTag['name'] . '</span>';
-        }
-
-        $res = implode(' ', $res);
-        break;
-
-      default:
-        $res = implode(', ', array_column($this->caseTags, 'name'));
+  private function getCaseTags() {
+    $res = [];
+    foreach ($this->caseTags as $caseTag) {
+      $res[] = '<span class="crm-tag-item" style="background-color: ' . $caseTag['background_color'] . '; color: ' . $caseTag['color'] . '">'
+        . $caseTag['name'] . '</span>';
     }
 
-    return $res;
+    return implode(' ', $res);
   }
 
   /**
@@ -460,7 +449,7 @@ class CRM_CiviAwards_Form_AwardReview extends CRM_Core_Form {
   private function getPageTitle() {
     $title = $this->caseContactDisplayName . ' - ' . $this->caseTypeName;
     if ($this->caseTags) {
-      $title = $title . ' &nbsp; ' . $this->getCaseTags('badge');
+      $title = $title . ' &nbsp; ' . $this->getCaseTags();
     }
 
     return $title;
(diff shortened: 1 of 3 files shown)


ang/civiawards/reviews-tab/directives/reviews-case-tab-content.directive.js
```diff
@@ -157,7 +159,29 @@
...
-      loadForm(formUrl);
+      loadForm(formUrl)
+        .on(CRM_FORM_LOAD_EVENT, popupTitleDecodeEntities);
+    }
+
+    /**
+     * Converts HTML entities in popup title to their corresponding characters.
+     */
+    function popupTitleDecodeEntities () {
+      $('.ui-dialog-title').each(function () {
```
reneolivo, line 170: We need to make sure that we only update the title for the modal form that we opened (multiple modals can be opened at the same time, although that's not very likely to happen in a real scenario).

I think we need to do something like: 

```js
// ...
$formElement = loadForm(formUrl)
        .on(CRM_FORM_LOAD_EVENT, popupTitleDecodeEntities);

// ...
$formElement.find('.ui-dialog-title').each(/* ... */)

```

ivan-compucorp: Fixed, thanks.

CRM/CiviAwards/Form/AwardReview.php
```diff
@@ -368,28 +368,88 @@ private function getCaseTypeName() {
...
+        $res = [];
+        foreach ($this->caseTags as $caseTag) {
+          $res[] = '<span class="crm-tag-item" style="background-color: ' . $caseTag['background_color'] . '; color: ' . $caseTag['color'] . '">' . $caseTag['name'] . '</span>';
+        }
+
+        $res = implode(' ', $res);
+        break;
+
+      default:
+        $res = implode(', ', array_column($this->caseTags, 'name'));
```
tunbola, line 423: Do we really need the default/plain format? Can't see it being called anywhere?
If not we can remove the switch and just make the function return  by default what the badge format returns.

CRM/CiviAwards/Form/AwardReview.php
```diff
@@ -368,28 +368,50 @@ private function getCaseTypeName() {
...
-    $caseTags = '';
+    $caseTags = [];
     foreach ($result['values'] as $caseTag) {
-      $caseTags .= $caseTag['api.Tag.getsingle']['name'] . ', ';
+      $caseTags[] = [
+        'id'               => $caseTag['api.Tag.getsingle']['id'],
+        'parent_id'        => $caseTag['api.Tag.getsingle']['parent_id'],
+        'name'             => $caseTag['api.Tag.getsingle']['name'],
+        'background_color' => !empty($caseTag['api.Tag.getsingle']['color']) ? $caseTag['api.Tag.getsingle']['color'] : '#ffffff',
+      ];
```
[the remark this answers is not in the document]
ivan-compucorp: Didn't understand - you suggest to remove this spacing or what?

CRM/CiviAwards/Form/AwardReview.php
```diff
@@ -408,7 +408,17 @@ private function loadCaseTags() {
    */
   private function getCaseTags($format = 'plain') {
     switch ($format) {
-      default: $res = implode(', ', array_column($this->caseTags, 'name'));
+      case 'badge':
+        $res = [];
+        foreach ($this->caseTags as $caseTag) {
+          $res[] = '<span class="crm-tag-item" style="background-color: ' . $caseTag['background_color'] . '">' . $caseTag['name'] . '</span>';
```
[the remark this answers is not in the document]
ivan-compucorp: Like this?
```
          $res[] = '<span class="crm-tag-item" style="background-color: ' . $caseTag['background_color'] . '">' .
            $caseTag['name'] . '</span>';
```

CRM/CiviAwards/Form/AwardReview.php
ivan-compucorp: There is a line 194 which uses 'plain' format:
```
    $this->assign('caseTags', $this->getCaseTags());
```

Not sure where this is used and could that be changed to use 'badge' format.
So we can make 'badge' format by default if you want, just want to confirm that we should.

ang/test/civiawards/reviews-tab/directives/reviews-case-tab-content.directive.spec.js
```diff
@@ -239,6 +239,43 @@
...
+        beforeEach(() => {
+          $scope.handleViewReviewActivity(selectedReview);
+        });
+
+        describe('after the form has been loaded', () => {
+          beforeEach(() => {
+            mockedFormElement.trigger(CRM_FORM_LOAD_EVENT);
+          });
+
+          it('case tags in popup title are html objects', () => {
```
deb1990, line 272: `it` statements should be a complete sentence. So something like
'displays the tags with background and font color'. So when you add `it`, then it sounds like `it displays the tags with background and font color`.

CRM/CiviAwards/Form/AwardReview.php
tunbola: Line 194 sets the `caseTags` to be used in the template but on checking `https://github.com/compucorp/uk.co.compucorp.civiawards/blob/0bce88d517410995dc5bbe1f450e28dd730c451d/templates/CRM/CiviAwards/Form/AwardReview.tpl` I can't seem to find that used anywhere. 

So it's safe to refactor the function.

ivan-compucorp: What should we do with line 194?
1. Remove it.
2. Use 'badge' format (which would be default after refactoring).
3. Use 'plain' format still.

tunbola: Remove line 194, refactor function to only return badge style, no need for function to accept any parameter since there are no other format.
Do we still need plain? It's not used anywhere

PR #105, opened 2020-06-22
RSE-1172: Award Review Form Changes for SSP

tunbola:
## Overview
This PR makes some changes to the Award review form for SSP.
The following changes are made:

- Changed the award review form url to use `civicrm/ssp/awardreview`
- Changed permission used by above URL to `CiviAwards: Access awards panel portal`
- Check that logged in contact belongs to a panel on the Award before being able to add a review.
- Fetch form prefix fields to be displayed on the award form fields.

## Before
- These changes were not yet done.

## After
- These changes were implemented.

When User tries to add a review for an award he is not panel member of.
Warning  CaseLatest 2020-06-22 12-01-29

### DIFF
--- a/CRM/CiviAwards/Form/AwardReview.php
+++ b/CRM/CiviAwards/Form/AwardReview.php
@@ -2,6 +2,8 @@
 
 use CRM_CiviAwards_ExtensionUtil as E;
 use CRM_CiviAwards_BAO_AwardDetail as AwardDetail;
+use CRM_CiviAwards_Service_AwardPanelContact as AwardPanelContact;
+use CRM_CiviAwards_Service_AwardApplicationContactAccess as AwardApplicationContactAccess;
 
 /**
  * Form controller class.
@@ -170,7 +172,36 @@ class CRM_CiviAwards_Form_AwardReview extends CRM_Core_Form {
       return 'You can only view the reviews that you have added!';
     }
 
+    if ($this->isReviewFromSsp() && !$this->contactHasPanelAccessToAward()) {
+      return 'You are not a member of any panel on this Award';
+    }
+
     return NULL;
+  }
+
+  /**
+   * Checks if the logged in contact has access to the Award.
+   *
+   * The contact has access if it belongs to a panel on the
+   * award.
+   *
+   * @return bool
+   *   Whether contact has access or not.
+   */
+  private function contactHasPanelAccessToAward() {
+    $loggedInContact = CRM_Core_Session::getLoggedInContactID();
+    $awardId = $this->getCaseTypeId();
+    $awardPanelContact = new AwardPanelContact();
+    $contactAccessService = new AwardApplicationContactAccess();
+
+    try {
+      $contactAccess = $contactAccessService->get($loggedInContact, $awardId, $awardPanelContact);
+
+      return !empty($contactAccess) ? TRUE : FALSE;
+    }
+    catch (Exception $e) {
+      return FALSE;
+    }
   }
 
   /**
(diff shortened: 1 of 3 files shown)


CRM/CiviAwards/Form/AwardReview.php
```diff
@@ -13,7 +15,7 @@ class CRM_CiviAwards_Form_AwardReview extends CRM_Core_Form {
   /**
    * URL for accessing review form from SSP.
    */
-  const SSP_REVIEW_URL = 'civicrm/award-review-ssp';
+  const SSP_REVIEW_URL = 'civicrm/ssp/awardreview';
```
swastikpareek, line 18: Not sure if we change this to `civicrm/ssp/award-review` for better readability. Leave it upto you.

CRM/CiviAwards/Form/AwardReview.php
```diff
@@ -170,9 +172,38 @@ private function getErrorMessage($fields) {
       return 'You can only view the reviews that you have added!';
     }
 
+    if ($this->isReviewFromSsp() && !$this->contactHasPanelAccessToAward()) {
+      return 'You are not a member of any panel on this Award';
```
swastikpareek, line 176: How multilingual support will work here? Just curious :)

templates/CRM/CiviAwards/Form/AwardReview.tpl
```diff
@@ -90,13 +90,13 @@
...
 
       {* Form fields section Starts *}
-      {foreach from=$elementNames item=elementName}
+      {foreach from=$elementData item=element}
         <div class="form-group {$form_group_class}">
-          <label class="{$form_group_label_class}">{$form.$elementName.label}</label>
+          <label class="{$form_group_label_class}">{$form[$element.name].label}</label>
           {if $isReviewFromSsp}
-            <div class="ssp-form-control-description text-muted"> Lorem ipsum dolor sit. Sit dolor ipsum lorem porem. </div>
+            <div class="ssp-form-control-description text-muted"> {$element.help_post} </div>
```
swastikpareek, line 97: TBH. I feel that this should be `<div class="ssp-form-control-description text-muted"> {$element.help_pre} </div>`

templates/CRM/CiviAwards/Form/AwardReview.tpl
```diff
@@ -90,13 +90,13 @@
...
           {/if}
-          <div class="{$form_group_field_class}">{$form.$elementName.html}</div>
+          <div class="{$form_group_field_class}">{$form[$element.name].html}</div>
```
swastikpareek, line 99: And `help_post` should be printed here
```
{if $isReviewFromSsp}
<div class="ssp-form-control-description text-muted"> {$element.help_post} </div>
```
{/if}

xml/Menu/civiawards.xml
```diff
@@ -7,10 +7,10 @@
     <access_arguments>access CiviCRM</access_arguments>
   </item>
   <item>
-    <path>civicrm/award-review-ssp</path>
+    <path>civicrm/ssp/awardreview</path>
```
swastikpareek, line 10: Same as above.

CRM/CiviAwards/Form/AwardReview.php
tunbola: We are supposed to use the ts function in the template. `{ts}` Some word {/ts}`

CRM/CiviAwards/Form/AwardReview.php
tunbola: I think this is fine because it's similar to the one for civicrm.

templates/CRM/CiviAwards/Form/AwardReview.tpl
tunbola: `Pre Help` are supposed to be help fields displayed before a form field in Civicrm. Post help are displayed in tooltips when the icon in front of the field is clicked. I think the choice is ours here since this is a bit custom. I've exposed both fields and we can use both as we deem fit.

xml/Menu/civiawards.xml
tunbola: This is fine as it's blends with URL for civicrm

templates/CRM/CiviAwards/Form/AwardReview.tpl
tunbola: To be honest, I think you can update the template as you deem fit. I didn't even want to update this. I only wanted to expose the variables.

swastikpareek: But, Isn't this feature part of this PR. Do you think we should have another PR just to add post and prefix. I think that's an overkill. If we have everything, why can't we just add it and be done with it?

swastikpareek: But, now I think, it's fine to keep it as it. As we don't know as of now which field will be used as post field. I think we can work on this as a part of the FE ticket as this might need discussion and it is not right to block this ticket/PR for that.

PR #111, opened 2020-07-03
RSE-1197: Add Applicant Management Instance Menu Class and Option Value

tunbola:
## Overview
This PR depends on https://github.com/compucorp/uk.co.compucorp.civicase/pull/515, The Applicant Management option value which is an option value of the Case Category Instance type option group was added in this PR. Also the menu class for the Applicant management Instance was added.

## Before
- The Applicant management Menu class and option value was not added.

## After
- The Applicant management Menu class and option value was added.

## Technical Details
- As explained in https://github.com/compucorp/uk.co.compucorp.civicase/pull/515, each Case category instance will have it's own menu class which will be responsible for creating menu Items for a case category belonging to a case category instance. The `ApplicantManagementMenu` class was added to create/update/delete menu items for case categories for this instance. The class uses most of the methods in the base `CRM_Civicase_Service_CaseCategoryMenu` class and only overrides the `createCaseCategorySubmenus` method.
- The Applicant Management option value was also added so that this option can be available on the Create case type category screen.

## Comments
PHP unit Tests are failing for Awards because the Awards extension expects the CaseCategory Entity to be present and this is not present on master and it is the master version of civicase is installed for test. There are some discussion around how to fix this and will be fixed in later PR's

### DIFF
--- a/CRM/CiviAwards/Helper/CaseTypeCategory.php
+++ b/CRM/CiviAwards/Helper/CaseTypeCategory.php
@@ -6,6 +6,7 @@
 class CRM_CiviAwards_Helper_CaseTypeCategory {
 
   const AWARDS_CASE_TYPE_CATEGORY_NAME = 'awards';
+  const APPLICATION_MANAGEMENT_NAME = 'applicant_management';
 
   /**
    * Returns the case types for the awards category.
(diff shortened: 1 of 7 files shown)


erawat:
@tunbola  Unit test fails when installing civiawards.
```

Enabling extension "uk.co.compucorp.civiawards"

                                                                    
  [Symfony\Component\Debug\Exception\FatalThrowableError]           
  Class 'CRM_Civicase_Setup_CaseCategoryInstanceSupport' not found  
                                                                    
```

PR #112, opened 2020-07-06
RSE-1140: Change Operator Between Applicant Status And Tags Field in Review Panel to `AND`

tunbola:
## Overview
Currently the operator between the case status and the case tags field when configuring a panel is the `OR` operator, i.e the panel member can view an application in a specific status or having a specific tag. However this is not correct as the specification states that the operator should be an `AND`. This PR makes that change.

Configure Award - Award Type  CaseLatest 2020-07-06 12-10-26

## Before
The operator between the Case Status and Case tags field is an `OR` 

## After
The operator between the Case Status and Case tags field is now an `AND`.
This change however brings some changes to the response of the  AwardReviewPanel.getContactAccess API.
Instead of returning an aggregated value for the application status and application tags for all panels, these values are now returned per panel since the application that a user has access to has to be fetched on a per panel basis.

```php
Sample Call and Response

$result = civicrm_api3('AwardReviewPanel', 'getcontactaccess', [
  'contact_id' => 7,
  'award_id' => 4,
]);

{

    "is_error": 0,
    "version": 3,
    "count": 3,
    "values": [
        {
            "application_tags": [],
            "application_status": [
                "1"
            ]
        },
        {
            "application_tags": [],
            "application_status": [
                "3"
            ]
        },
        {
            "application_tags": [
                "7"
            ],
            "application_status": [
                "3"
            ]
        }
    ]
}


```

The `AwardReveiewPanel.getReviewAcess` request and response remains the same.

### DIFF
--- a/tests/phpunit/CRM/CiviAwards/Service/AwardApplicationContactAccessTest.php
+++ b/tests/phpunit/CRM/CiviAwards/Service/AwardApplicationContactAccessTest.php
@@ -93,39 +93,6 @@ class CRM_CiviAwards_Service_AwardApplicationContactAccessTest extends BaseHeadl
           'is_application_status_restricted' => 0,
         ],
       ],
-    ];
-    $awardPanel = [];
-    foreach ($params as $param) {
-      $awardPanel[] = AwardReviewPanelFabricator::fabricate($param);
-    }
-
-    $awardPanelContact = $this->getAwardPanelContactObject($awardPanel, $contactId);
-    // Anonymize application will be false overral because more permission takes
-    // precedence over less less permissions.
-    $expectedResult = [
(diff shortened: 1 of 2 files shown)


tests/phpunit/CRM/CiviAwards/Service/AwardApplicationContactAccessTest.php
```diff
@@ -336,13 +265,16 @@ public function testGetReviewAccessForContactAndCaseHasTags() {
   }
 
   /**
-   * Tes When Case Has No Tags And Contact Cannot View Anonymised.
+   * Tes When Panel Setting Is For All Tags And Contact Cannot View Anonymised.
```
hitesh-jain, line 268: @tunbola typo "Tes"

PR #114, opened 2020-07-07
RSE-1219: Text changes on the SSP awards form pages

swastikpareek:
## Overview
As a part of this PR, text and label changes are done on the awards pages

#### Before
Screenshot 2020-07-07 at 7 33 04 PM

#### After
Screenshot 2020-07-07 at 7 31 39 PM

#### Before
title issue - before

#### After
title issue - after

#### Before
Screenshot 2020-07-07 at 9 18 41 PM

#### After
Screenshot 2020-07-07 at 9 15 35 PM

## Technical Details
* Change text in the `AwardReview.php` to match the design - https://projects.invisionapp.com/d/main#/console/19802243/418400657/preview
* Change the headline for SSP pages to use case contact name from source contact name.
* Change error message text in `AwardReview.php` as per the implementation.

### DIFF
--- a/templates/CRM/CiviAwards/Form/AwardReview.tpl
+++ b/templates/CRM/CiviAwards/Form/AwardReview.tpl
@@ -39,11 +39,7 @@
         <div class="form-group {$form_group_class}">
           {if $isReviewFromSsp}
             <h1 class="header-pager">
-              {if $isViewAction}
-                {$sourceContactName}
-              {else}
-                {$form.source_contact_id.html}
-              {/if}
+              {$caseContactDisplayName}
             </h1>
             <div class="ssp-details-page-section__sub-heading">
               <span class="ssp-applicant-card__award ssp-text-large">{$caseTypeName}</span>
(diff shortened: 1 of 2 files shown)


templates/CRM/CiviAwards/Form/AwardReview.tpl
```diff
@@ -39,11 +39,7 @@
         <div class="form-group {$form_group_class}">
           {if $isReviewFromSsp}
             <h1 class="header-pager">
-              {if $isViewAction}
-                {$sourceContactName}
-              {else}
-                {$form.source_contact_id.html}
-              {/if}
+              {$caseContactDisplayName}
```
tunbola, line 42: We don't need the else part of the if statement above?

swastikpareek: Don't think so. We need to show the same contact display name to view and non-view pages. Right?

swastikpareek: @tunbola

tunbola: The logic above shows that on Non view pages, i.e on create, A form field should show. Can you confirm create and view works fine after this change?

swastikpareek: Yeas. Both the view page and form page works fine. 
Screenshot 2020-07-07 at 9 23 09 PM
Screenshot 2020-07-07 at 9 23 52 PM

PR #119, opened 2020-07-13
RSE-1200: Custom Field Support For Applicant Management Case Types

tunbola:
## Overview
Custom field support for Award Category case types was already done in https://github.com/compucorp/uk.co.compucorp.civiawards/pull/94. This PR makes some changes to make this to support Applicant Management case types rather than just Award Case Types. This means that Case types for all case categories belonging to the Applicant Management instance can support having custom fields.

## Before
These changes were not present

## After
- Some classes were renamed to reflect Applicant Management rather than Awards case category.
- Most importantly, the `getAwardCaseType` function in the CaseCategory Helper was renamed to `getApplicantManagementCaseTypes` and this function now returns case types under the Applicant management instance.

## Comments
- An issue experience when installing Civicase and CiviAwards for the first time was also fixed, The CaseCategory Entity recently added in the CiviCase extension will require all caches to be cleared before it becomes available to be used as an API on first install of the Civicase extension, The BAO is simply used in the `CRM_CiviAwards_Setup_CreateApplicantManagementOption` class for the update here so that clearing the whole cache doesn't slow the installation process.
- The Github linter Workflow was updated not to run linters for deleted files. Same thing was done [here](https://github.com/compucorp/uk.co.compucorp.civicase/commit/420dd53e69e22c6f24ac9d625a6c22f6948cbe86) for CiviCase


.github/workflows/linters.yml
```diff
@@ -20,17 +20,17 @@ jobs:
...
       #the remaining step will not be executed. 
       if: ${{ always() }} 
-      run: git diff origin/${GITHUB_BASE_REF} --name-only -- '*.scss' | xargs -r  npx stylelint --config .stylelintrc 
+      run: git diff --diff-filter=d origin/${GITHUB_BASE_REF} --name-only -- '*.scss' | xargs -r  npx stylelint --config .stylelintrc
         
 
     - name: Run eslint linter
       if: ${{ always() }}
-      run: git diff origin/${GITHUB_BASE_REF} --name-only -- '*.js' | xargs -r  npx eslint --max-warnings=0 -c .eslintrc.json 
+      run: git diff  --diff-filter=d origin/${GITHUB_BASE_REF} --name-only -- '*.js' | xargs -r  npx eslint --max-warnings=0 -c .eslintrc.json
```
reneolivo, line 36: @tunbola this removed the last empty line from the file.

PR #121, opened 2020-07-14
RSE-1199: Add AwardDetail.getInstanceCategories API

tunbola:
## Overview
This PR adds an API that will return the case categories that are of the Applicant Management Category Instance.

## Before
- This API is not present

## After
The AwardDetail.getInstanceCategories is added

```php
Sample Request and Response
civicrm_api3('AwardDetail', 'getinstancecategories');

{

    "is_error": 0,
    "version": 3,
    "count": 2,
    "values": {
        "2": "Awards",
        "3": "Awards Project"
    }
}

```

### DIFF
--- a/CRM/CiviAwards/Test/Fabricator/CaseCategory.php
+++ b/CRM/CiviAwards/Test/Fabricator/CaseCategory.php
@@ -1,0 +1,41 @@
+<?php
+
+/**
+ * Class CRM_CiviAwards_Test_Fabricator_CaseCategory.
+ */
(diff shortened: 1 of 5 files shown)


CRM/CiviAwards/Test/Fabricator/CaseCategory.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+/**
+ * Class CRM_CiviAwards_Test_CaseCategory.
```
hitesh-jain, line 4: @tunbola looks like typo here in class name

PR #123, opened 2020-08-10
DRU-190: Show error message for ssp awards review application pages where application is deleted

hitesh-jain:
## Overview

This PR fixes issue with SSP Awards Review application page where if the review panel member has direct URL of review application page or view submitted review page they can submit/view reviews for the applications which were in deleted state but we should rather show an error message "You cannot submit review for a deleted application." on add review page and "You cannot view review for a deleted application." on view submitted review page.

## Before
Case application is deleted from backend but if review panel member has direct URL they can access and add reviews.
image

View submitted review page
image

## After
image

image

## Technical Details

- Added check for Case application "is_deleted" field for SSP review forms and return error message.

### DIFF
--- a/CRM/CiviAwards/Form/AwardReview.php
+++ b/CRM/CiviAwards/Form/AwardReview.php
@@ -170,6 +170,11 @@ class CRM_CiviAwards_Form_AwardReview extends CRM_Core_Form {
     $hasSubmittedReview = $this->userAlreadySubmittedReview();
     $canNotViewReview = $isViewAction && $this->isReviewFromSsp() && !$this->isReviewOwner();
 
+    if ($this->isReviewFromSsp() && $this->isCaseApplicationDeleted()) {
+      $action = $isViewAction ? 'view' : 'submit';
+      return "You cannot $action a review for a deleted application.";
+    }
+
     if ($this->isReviewFromSsp() && $hasSubmittedReview && $isAddAction) {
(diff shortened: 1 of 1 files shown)


CRM/CiviAwards/Form/AwardReview.php
```diff
@@ -125,6 +125,9 @@ public function preProcess() {
       $this->caseId = CRM_Utils_Array::first($this->activity['case_id']);
     }
 
+    if ($this->isCaseApplicationDeleted()) {
+      drupal_not_found();
```
tunbola, line 129: Since Civicase is a civicrm extension and it's possible that the extension can be used with Civicrm in combination with another CMS like wordpress or Joomla. We should not have drupal specific code here. 

I would suggest we move the logic to check deleted applications to the getErrorMessage function and return the appropriate error message there.  See : https://github.com/compucorp/uk.co.compucorp.civiawards/blob/DRU-190-deleted-applications-access/CRM/CiviAwards/Form/AwardReview.php#L164-L189

This will also be consistent with previous error message display implementation.

tunbola: Also we need to check if we want this behaviour on the Civicrm side. If not, we can add the `$this->isReviewFromSsp()` in addition to the application deleted condition if we need to restrict to SSP only.

hitesh-jain: fixed

PR #125, opened 2020-08-24
RSE-1248: Fix bug in getreviewaccessapi

shahrukh-compucorp:
## Overview
The applications which do not have any tag attached to them are now visible to review panel users if panel does not contain any tag.

## Before
If there is a review panel attached to an award that does not include any tags then the user belonging to that panel did not have visibility of the applications which do not have any tag attached to it.

## After
If there is a review panel attached to an award that does not include any tags then the user belonging to that panel is now able to see all the applications regardless of whether the application contains any tags or not.

## Technical Details
The logic that was responsible for checking the tags matching had a glitch in it. The condition was not catering the situation where application has empty tags. So the logic was changed to if there are no tags defined in the review panel then the applications without any tags will be accessible to the user

### DIFF
--- a/CRM/CiviAwards/Service/AwardApplicationContactAccess.php
+++ b/CRM/CiviAwards/Service/AwardApplicationContactAccess.php
@@ -138,7 +138,7 @@ class CRM_CiviAwards_Service_AwardApplicationContactAccess {
 
     foreach ($visibilitySettings as $visibilitySetting) {
       $caseStatusMatch = array_intersect($visibilitySetting['application_status'], $caseStatus) || empty($visibilitySetting['application_status']);
-      $caseTagMatch = (empty($visibilitySetting['application_tags']) || (!empty($caseTags) && array_intersect($visibilitySetting['application_tags'], $caseTags)));
+      $caseTagMatch = empty($visibilitySetting['application_tags']) || (!empty($caseTags) && array_intersect($visibilitySetting['application_tags'], $caseTags));
 
       if ($caseStatusMatch && $caseTagMatch) {
         $statusToMoveApplicationTo = array_merge($statusToMoveApplicationTo, $visibilitySetting['status_to_move_to']);
--- a/CRM/CiviAwards/Service/AwardApplicationContactAccess.php
+++ b/CRM/CiviAwards/Service/AwardApplicationContactAccess.php
@@ -138,7 +138,7 @@ class CRM_CiviAwards_Service_AwardApplicationContactAccess {
 
     foreach ($visibilitySettings as $visibilitySetting) {
       $caseStatusMatch = array_intersect($visibilitySetting['application_status'], $caseStatus) || empty($visibilitySetting['application_status']);
-      $caseTagMatch = !empty($caseTags) && (array_intersect($visibilitySetting['application_tags'], $caseTags) || empty($visibilitySetting['application_tags']));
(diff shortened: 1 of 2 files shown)


CRM/CiviAwards/Service/AwardApplicationContactAccess.php
```diff
@@ -138,7 +138,7 @@ public function getReviewAccess($contactId, $applicationId, AwardReviewPanelCont
 
     foreach ($visibilitySettings as $visibilitySetting) {
       $caseStatusMatch = array_intersect($visibilitySetting['application_status'], $caseStatus) || empty($visibilitySetting['application_status']);
-      $caseTagMatch = !empty($caseTags) && (array_intersect($visibilitySetting['application_tags'], $caseTags) || empty($visibilitySetting['application_tags']));
+      $caseTagMatch = (empty($visibilitySetting['application_tags']) || (!empty($caseTags) && array_intersect($visibilitySetting['application_tags'], $caseTags)));
```
tunbola, line 141: We really don't need the first opening curly bracket and the last curly brackets `(` `)`

tests/phpunit/CRM/CiviAwards/Service/AwardApplicationContactAccessTest.php
```diff
@@ -425,8 +425,8 @@ public function testContactIsNotAbleToMoveToAnyStatusWhenCaseHasNoTags() {
     // `AND` operator between the status and case tags, it is compulsory for
     // a case to have at least a tag.
     $expectedResult = [
```
tunbola, line 427: We need to update the comment here and also update the test name since the test name no longer reflects what is being happening in the test.

A suggestion might be `testContactIsAbleToMoveToRelevantStatusWhenCaseHasNoTagsAndPanelTagsIsEmpty` or some other better name.

tunbola (changes requested):

tunbola:
@shahrukh-compucorp , A few comments
- Can you remove the line before the Overview section in the description.
- The overview section is actually supposed to describe the problem in lay man's terms, keeping it non-technical
- You can add a bit more details in the after section or use the Technical Section details to explain what was actually wrong technically with the existing logic and how it was fixed, the PR currently does not describe what was technically wrong
For some PR's you can add GIF's in the before and after section to explain the issue but I don't think that will be useful for this particular PR. Just a note.

CRM/CiviAwards/Service/AwardApplicationContactAccess.php
```diff
@@ -138,7 +138,7 @@ public function getReviewAccess($contactId, $applicationId, AwardReviewPanelCont
 
     foreach ($visibilitySettings as $visibilitySetting) {
       $caseStatusMatch = array_intersect($visibilitySetting['application_status'], $caseStatus) || empty($visibilitySetting['application_status']);
-      $caseTagMatch = (empty($visibilitySetting['application_tags']) || (!empty($caseTags) && array_intersect($visibilitySetting['application_tags'], $caseTags)));
+      $caseTagMatch = empty($visibilitySetting['application_tags']) || (!empty($caseTags) && array_intersect($visibilitySetting['application_tags'], $caseTags));
```
tunbola, line 141: The indentation here seems off. Can you fix?

tunbola (changes requested):

tunbola (approved):

tunbola:
@shahrukh-compucorp , can you also fix commit author for this PR.

tunbola (changes requested):
@shahrukh-compucorp , can you also fix commit author for this PR.

tunbola (approved):

PR #126, opened 2020-08-25
RSE-1226: Disable fields for review in view mode

shahrukh-compucorp:
## Overview
When a user opens a review in view mode all fields should be disabled as it did not make sense to let the user edit the fields when there is no option for saving the form in view mode

## Before
Text and select fields were already disabled but attachment, radio, checkboxes fields were not disabled
8d3bb2ce-ec8a-4f46-9584-8ebcd3cdeee9

## After
No matter what the type of a field is, it will be disabled in view mode
Screenshot from 2020-08-25 17-51-20

## Technical Details
Freeze function was being called for form fields in view mode to disable the form fields which was not supported by radio, checkbox and file fields so in addition to the freeze method the disabled attribute is also added for each form field's attributes list in view mode

### DIFF
--- a/CRM/CiviAwards/Form/AwardReview.php
+++ b/CRM/CiviAwards/Form/AwardReview.php
@@ -318,9 +318,12 @@ class CRM_CiviAwards_Form_AwardReview extends CRM_Core_Form {
     $elementNames = array_keys($elementData);
     if ($isViewAction) {
       foreach ($this->_elements as $element) {
+        // This check eliminates checkbox and radio buttons as
+        // freeze should not be called on those elements.
         if (in_array($element->_attributes['name'], $elementNames)) {
           $element->freeze();
         }
+        $this->disableField($element);
       }
     }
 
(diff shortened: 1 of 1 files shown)


CRM/CiviAwards/Form/AwardReview.php
```diff
@@ -321,6 +321,7 @@ private function displayReviewForm(array $fields) {
         if (in_array($element->_attributes['name'], $elementNames)) {
           $element->freeze();
         }
+        $this->disableField($element);
```
tunbola, line 324: Any reason why don't we move this inside the `if` statement above ? We should only disable elements that are present in the `$elementNames` array I think.

tunbola (changes requested):

CRM/CiviAwards/Form/AwardReview.php
shahrukh-compucorp: well some elements does not have name field in attributes so that is why placed it outside the if and tbh there are only some extra hidden fields and I think it wont cause any issue if I disable them in view  mode. Let me know your thoughts on it

tunbola:
@shahrukh-compucorp , can you rebase commits. The commit author is wrong. Thanks

CRM/CiviAwards/Form/AwardReview.php
tunbola: @shahrukh-compucorp , it's okay but the logic needs to be consistent. If we are freezing only some elements and disabling some, then there is inconsistency. It's either we disable and freeze all and remove the IF statement or we only disable and freeze those in the IF statement.

That said will suggest we move the disable function into the IF statement, since we are sure that those are the elements we want to freeze/disable.

tunbola (approved):

PR #128, opened 2020-09-10
RSE-1299: Fix contact display name on SSP review application form to respect anonymization

hitesh-jain:
## Overview

This PR implements Anonymizing Contact Display name on SSP review case application form and view submitted case application form where if the awards review panel is marked with "Anonymise Contacts for this Panel" we shouldn't show Contact Display name of that case application instead we need to show "Anonymous [Case ID]" as display name to awards review panel member while reviewing that case application.

## Before
image

image

## After
image

## Technical Details

- Add check for SSP review application form to check if that case application has "Anonymize Application" param to true and if yes return "Anonymous [case-id]" as display name


tunbola (approved):

tunbola:
@hitesh-jain can you add after screenshot?

PR #132, opened 2020-09-21
RSE-1246: Disallow empty statuses if Panel can change the status of an application checkbox is ticked

shahrukh-compucorp:
## Overview
Statuses field should be filled in to allow panel users to move the application between statuses if 'Panel can change the status of an application' checkbox is ticked.

## Before
User was able to save the review panel with empty statuses field and 'Panel can change the status of an application' checkbox ticked
Screenshot from 2020-09-21 20-27-50

## After
If user ticks 'Panel can change the status of an application' checkbox then he should select atleast one status otherwise review panel will not be saved
Screenshot from 2020-09-21 20-20-28

## Technical Details
The fix for this required checks to be added on both backend and frontend as well. On the frontend a conditional required was added to the status field which will not allow the user to submit the form until he select atleast one status if 'Panel can change the status of an application' checkbox is ticked. On the backend the logic for validating input was changed as previously if a field was required the code only checked for its presence which did not seem to work for array fields so now if a field is required and it is of type array then it must contain atleast one value.
https://github.com/compucorp/uk.co.compucorp.civiawards/blob/1942cd62ae751a51f97ed2e176cf06275f1a088b/CRM/CiviAwards/BAO/AwardReviewPanel.php#L181


reneolivo (dismissed):

tunbola (changes requested):
@shahrukh-compucorp , Can you please add a BE test to cover this change. Thanks

tests/phpunit/CRM/CiviAwards/BAO/AwardReviewPanelTest.php
```diff
@@ -216,12 +254,37 @@ public function testCreateThrowsExceptionWhenRestrictedStatusIsEmptyAndApplicati
     AwardReviewPanelFabricator::fabricate($params);
   }
 
+  /**
+   * Test create function fails if restricted status is empty.
+   */
+  public function testCreateThrowsExceptionWhenRestrictedStatusIsEmptyAndApplicationStatusIsRestricted() {
```
tunbola, line 260: Will be a good idea to separate syntax fixes commit from other code commits next time so it's easier to review the PR.

tunbola:
@shahrukh-compucorp , 
- Can you please fix the PR title, It's truncated.

tunbola (changes requested):

tunbola (approved):

tests/phpunit/CRM/CiviAwards/BAO/AwardReviewPanelTest.php
shahrukh-compucorp: sure

tunbola (approved):

PR #130, opened 2020-09-14
RSE-1296: Award Custom Field UI

deb1990:
## Overview
As part of this PR, the UI to save/edit Custom Fields for an Award is added.

## Before
The feature was not present.

## After
after

## Technical Details
1. Added `<civiaward-custom-field-sets></civiaward-custom-field-sets>` directive. Which handles this feature.
2. In `award.directive.js`, added `saveIndividualTabsIfApplicable` function, which checks and waits for individual tabs save logic if present.
3. `templates/CRM/CiviAwards/Form/AwardCustomField.tpl` files indentation was wrong, fixed it.

### Update
Fixed 2 bugs
1. When creating Awards, the custom fields should not be fetched.
2. When there are no custom fields present for an award, we should see an empty message.

2020-10-01 at 4 57 PM

PR #138, opened 2020-10-13
CPS-360: Fix save award bug when there are no custom fields defined

reneolivo:
## Overview
This PR fixes an issue that happens when trying to save award types, but there are no custom fields defined.

## Before
gif

### When there are no custom fields
gif

### When there are custom fields
gif

## Technical Details

This is a regression issue introduced by https://github.com/compucorp/uk.co.compucorp.civiawards/pull/130 specifically this commit: https://github.com/compucorp/uk.co.compucorp.civiawards/commit/77b99e2

The problem is that when no custom fields are defined, the award form will wait for the non-existent form to trigger a success message, but this will never happen.

To fix this, we automatically resolve the save process of the award's custom field set directive if there are no custom fields defined.

### DIFF
--- a/ang/civiawards/award-creation/directives/award-custom-field-sets.directive.js
+++ b/ang/civiawards/award-creation/directives/award-custom-field-sets.directive.js
@@ -43,20 +43,22 @@
 
       /**
        * Save the custom field set PHP form by triggering the hidden save button
-       * using jQuery
+       * using jQuery.
+       *
+       * If there is no custom fields form, it will continue as normal.
        *
        * @returns {Promise} promise
        */
       function save () {
+        var $submitButtonLink = $('.civiaward__custom-field-sets__container .award-custom-field');
+        var hasCustomFields = $submitButtonLink.length > 0;
         var defer = $q.defer();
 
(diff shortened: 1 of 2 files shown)


jamienovick:
Small point - I don't think we should show the custom fields tab if there are no custom fields!

reneolivo:
@jamienovick, I think that's a question we need to ask the ux experts, but I'd say it's best to show it.

To me, it would feel like hiding a feature from the user and confusing them. If anything, I'd even add a link so they can create the custom fields if none exist. But that's just me. 

Also, this is a bug fixing PR, so better no introduce more features or we risk adding more bugs 😅

jamienovick:
UX preferances aside that's how it works everywhere else in CiviCRM. Remember also that not all users:

A) have access to create custom fields
B) have access to view all custom field sets

So there are a lot of situations where it could be blank and the create link is only relevant to admin (but I agree with the concept let's say for other future products!).

I think it will be important to hide this if it's empty at some point in future so we are aligned to how the rest of Civi works...

deb1990 (approved):

PR #145, opened 2020-10-28
RSE-1455: Fix Duplicate Workflow functionality

deb1990:
## Overview
Fixed the Duplicate `applicant_management` workflow feature. 

## Technical Details
1. Created a new angularjs module `civiawards-workflow`, which is dependent on `workflow`.
2. Added `DuplicateWorkflowApplicantmanagement`, with the logic to duplicate applicant_management workflows.

### DIFF
--- a/ang/civiawards-workflow/duplicate-workflow-applicant-management.service.js
+++ b/ang/civiawards-workflow/duplicate-workflow-applicant-management.service.js
@@ -18,7 +18,7 @@
     function create (workflow) {
       return civicaseCrmApi([
         ['AwardDetail', 'get', { sequential: 1, case_type_id: workflow.id }],
-        ['CaseType', 'create', _.extend(workflow, { id: null })]
+        ['CaseType', 'create', _.extend({}, workflow, { id: null })]
       ]).then(function (data) {
         var awardsDetailsData = data[0].values[0];
 
--- a/ang/civiawards-workflow/duplicate-workflow-applicant-management.service.js
+++ b/ang/civiawards-workflow/duplicate-workflow-applicant-management.service.js
@@ -1,0 +1,34 @@
+(function (_, angular) {
+  var module = angular.module('civiawards-workflow');
+
+  module.service('DuplicateWorkflowApplicantmanagement', DuplicateWorkflow);
+
+  /**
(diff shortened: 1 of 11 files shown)


ang/civiawards-workflow/duplicate-workflow-applicant-management.service.js
```diff
@@ -0,0 +1,34 @@
...
+    this.create = create;
+
+    /**
+     * @param {object} workflow workflow object
+     * @returns {Promise} promise
+     */
+    function create (workflow) {
+      return civicaseCrmApi([
+        ['AwardDetail', 'get', { sequential: 1, case_type_id: workflow.id }],
+        ['CaseType', 'create', _.extend(workflow, { id: null })]
```
reneolivo, line 21:
```js
_.extend({}, workflow, { id: null })
```

civiawards.php
```diff
@@ -242,6 +242,13 @@ function civiawards_addCiviCaseDependentAngularModules(&$dependentModules) {
   $dependentModules[] = "civiawards";
 }
 
+/**
+ * Implements addCiviCaseDependentAngularModules().
+ */
+function civiawards_addWorkflowDependentAngularModules(&$dependentModules) {
+  $dependentModules[] = "civiawards-workflow";
```
reneolivo, line 249: @deb1990 I understand that there was a new hook added to civicase to support this, but what's the real purpose?

Why not add this dependency directly to civiawards?

reneolivo (changes requested):

civiawards.php
deb1990: @reneolivo Right, because we dont want to add all the awards dependency to this page, because its going to bloat this workflows page. Its the same argument for creating the `workflows` module in civicase.

reneolivo: Why not create a new angular page using the `civiawards-workflow` angular module and this module includes `civicase-base` and `civiawards` ?

This hook pattern seems to be more difficult to understand than the standard process.

deb1990: This was discussed, and we agreed to keep this as it is.

reneolivo (approved):

PR #164, opened 2020-12-24
CPS-411: Import awards using CSV importer extension

lisandro-compucorp:
## Overview
This PR adds the functionality of mass-importing Awards, using the [CSV importer extension](https://github.com/eileenmcnaughton/nz.co.fuzion.csvimport) and a new "helper" entity named "AwardImport".

**UPDATE**
It was added improvements on the error handling, for the situations in which some of these fields have an empty value: Award Title, Award Start Date, Award Subtype, Award Manager.
More explanation about these changes on the `Technical Details` part of this PR.

## Before
Before there was not possible to import the Awards using the mentioned extension, because it has some related entities that couldn't be imported at the same time. 
The related entities are `AwardDetails`, `AwardManager`, `AwardReviewPanel`, and some other.

## After
Now, going to the CSV Importer extension ( Administer > API csv import ), on the first screen of the process, the `AwardImport` entity can be selected.
image
And in the second step, the fields for Awards can be mapped accordingly:
image

The new entity provides fields for loading all the basic and additional data for the awards.
Can be specified the information for: 
- case type (title, description)
- award details (start_date, end_date, is_active, is_template) 
- award manager (comma-separated list of contacts id)

## Comments
I am attaching an example CSV file.

[awards.txt](https://github.com/compucorp/uk.co.compucorp.civiawards/files/5815915/awards.txt)

## Technical Details

All the error handling was being delegated to the respective entities, for example, with award details:
```php
try {
      civicrm_api3('AwardDetail', 'create', $params);
    }
    catch (Exception $exception) {
      $tx->rollback();

      throw new API_Exception('Exception while saving the AwardDetail: ' . $exception->getMessage());
    }
```
But this was not considering (or at least informing correctly) the cases in which the value was empty. This is done on the frontend part in the current paths (outside the CSV importer)

For this reason, 4 new commits were added:

1. https://github.com/compucorp/uk.co.compucorp.civiawards/pull/164/commits/3b0a689326dc2e1f0ae9bc577c44aaed4482b05b For the empty Award Title (related to Case Type Title)
For this information we don't have the entity on the package (`CaseType`) then I added the handling of the particular error on the Service itself.

2. https://github.com/compucorp/uk.co.compucorp.civiawards/pull/164/commits/dfb78f3d3ea956d41659d06f60579328b5df7389 for the empty Award Subtype.
Inside the award details BAO.

3. https://github.com/compucorp/uk.co.compucorp.civiawards/pull/164/commits/7b9440e8f2bd4daca75e55b9f44e069d2a94181d for the empty Award Start Date.
Also inside the AwardDetail BAO.

4. https://github.com/compucorp/uk.co.compucorp.civiawards/pull/164/commits/61cc6be43d8516af12cc8d570336c8f47aeb1e54 for the empty Award Manager.
In a new method of the AwardDetail BAO, `validateAwardManager`, playing a similar role than the already existing `validateReviewFields`.
Also, the code iterates over the list looking for the contact and giving information when it is not found.

I also manually tested the upload, and do the normal path of Award Creation.

**Related to informing the row with the error.**
There is no easy way to know the row that failed, but the extension is printing all the row with the error, along with the error received. Since the Award Title is unique, the row can be identification easily.
Example:
This row:
>"test award - TD5 - 1","description award imported 1",1,2,,,0,2

Will produce this error:

>"Error with entity ""create""! (Exception while saving the AwardDetail: Award Start Date should not be empty)","test award - TD5 - 1","description award imported 1",1,2,,,0,2

(notice that is the error + the full row)

### DIFF
--- a/CRM/CiviAwards/BAO/AwardDetail.php
+++ b/CRM/CiviAwards/BAO/AwardDetail.php
@@ -81,6 +81,10 @@ class CRM_CiviAwards_BAO_AwardDetail extends CRM_CiviAwards_DAO_AwardDetail {
    *   Parameters.
    */
   private static function validateDates(array $params) {
+    if (empty($params['start_date'])) {
+      throw new Exception("Award Start Date should not be empty");
+    }
+
     if (!empty($params['start_date']) && !empty($params['end_date'])) {
       $startDate = new DateTime($params['start_date']);
       $endDate = new DateTime($params['end_date']);
--- a/CRM/CiviAwards/BAO/AwardDetail.php
+++ b/CRM/CiviAwards/BAO/AwardDetail.php
@@ -66,13 +66,30 @@ class CRM_CiviAwards_BAO_AwardDetail extends CRM_CiviAwards_DAO_AwardDetail {
    *   Parameters.
    */
   private static function validateParams(array &$params) {
-    if (empty($params['award_subtype'])) {
-      throw new Exception('Award Subtype should not be empty');
-    }
-
+    self::validateMandatory($params);
     self::validateDates($params);
     self::validateReviewFields($params);
     self::validateAwardManager($params);
+  }
+
+  /**
+   * Ensures that required fields are provided.
+   *
+   * @param array $params
+   *   Parameters.
+   */
+  private static function validateMandatory(array $params) {
+    $mandatoryFields = [
+      'start_date' => 'Award Start Date',
+      'award_subtype' => 'Award Subtype',
+      'award_manager' => 'Award Manager',
+    ];
+
+    foreach ($mandatoryFields as $field => $label) {
+      if (empty($params[$field])) {
+        throw new Exception("The {$label} field should not be empty");
+      }
+    }
   }
 
   /**
@@ -82,10 +99,6 @@ class CRM_CiviAwards_BAO_AwardDetail extends CRM_CiviAwards_DAO_AwardDetail {
    *   Parameters.
    */
   private static function validateDates(array $params) {
-    if (empty($params['start_date'])) {
-      throw new Exception("Award Start Date should not be empty");
-    }
-
     if (!empty($params['start_date']) && !empty($params['end_date'])) {
       $startDate = new DateTime($params['start_date']);
       $endDate = new DateTime($params['end_date']);
@@ -129,10 +142,6 @@ class CRM_CiviAwards_BAO_AwardDetail extends CRM_CiviAwards_DAO_AwardDetail {
    * @throws \Exception
    */
   private static function validateAwardManager(array $params) {
-    if (empty($params['award_manager'])) {
-      throw new Exception('Award Manager should not be empty');
-    }
-
     $contactManagers = (array) $params['award_manager'];
     foreach ($contactManagers as $contactId) {
       $contact = civicrm_api3('Contact', 'get', [
--- a/CRM/CiviAwards/BAO/AwardDetail.php
+++ b/CRM/CiviAwards/BAO/AwardDetail.php
@@ -72,6 +72,7 @@ class CRM_CiviAwards_BAO_AwardDetail extends CRM_CiviAwards_DAO_AwardDetail {
 
     self::validateDates($params);
     self::validateReviewFields($params);
+    self::validateAwardManager($params);
   }
 
   /**
@@ -117,6 +118,30 @@ class CRM_CiviAwards_BAO_AwardDetail extends CRM_CiviAwards_DAO_AwardDetail {
     $reviewFields = CRM_Utils_Array::value('review_fields', $params);
     $reviewFieldsValidator = new CRM_CiviAwards_Helper_CreateReviewFieldsValidator();
     $reviewFieldsValidator->validate($reviewFields);
+  }
+
+  /**
+   * Validates award_manager options to ensure legal options are passed.
+   *
+   * @param array $params
+   *   Request Parameters.
+   *
+   * @throws \Exception
+   */
+  private static function validateAwardManager(array $params) {
+    if (empty($params['award_manager'])) {
+      throw new Exception('Award Manager should not be empty');
+    }
+
+    $contactManagers = (array) $params['award_manager'];
+    foreach ($contactManagers as $contactId) {
+      $contact = civicrm_api3('Contact', 'get', [
+        'id' => $contactId,
+      ]);
+      if ($contact['count'] == 0) {
+        throw new Exception("Invalid Contact Received: {$contactId}");
+      }
+    }
   }
 
 }
--- a/CRM/CiviAwards/BAO/AwardDetail.php
+++ b/CRM/CiviAwards/BAO/AwardDetail.php
@@ -66,6 +66,10 @@ class CRM_CiviAwards_BAO_AwardDetail extends CRM_CiviAwards_DAO_AwardDetail {
    *   Parameters.
    */
   private static function validateParams(array &$params) {
+    if (empty($params['award_subtype'])) {
+      throw new Exception('Award Subtype should not be empty');
+    }
+
     self::validateDates($params);
     self::validateReviewFields($params);
   }
--- a/CRM/CiviAwards/Service/AwardImport.php
+++ b/CRM/CiviAwards/Service/AwardImport.php
@@ -29,8 +29,7 @@ class CRM_CiviAwards_Service_AwardImport {
 
     $params['case_type_id'] = $caseType['id'];
     $params['award_manager'] = !empty($params['award_manager']) ? explode(',', $params['award_manager']) : [];
-
-    $params['review_fields'] = !empty($params['review_fields']) ? json_decode($params['review_fields'], TRUE) : [];
+    $params['review_fields'] = [];
     try {
       civicrm_api3('AwardDetail', 'create', $params);
     }
--- a/CRM/CiviAwards/Service/AwardImport.php
+++ b/CRM/CiviAwards/Service/AwardImport.php
@@ -16,7 +16,7 @@ class CRM_CiviAwards_Service_AwardImport {
   public function create(array $params) {
     $tx = new CRM_Core_Transaction();
 
-    $params['case_type_category'] = $this->getCaseCategoryValueForAwards();
+    $params['case_type_category'] = 'awards';
     $params['definition'] = $this->createDefinition($params);
     try {
       $caseType = civicrm_api3('CaseType', 'create', $params);
--- a/CRM/CiviAwards/Service/AwardImport.php
+++ b/CRM/CiviAwards/Service/AwardImport.php
@@ -12,6 +12,8 @@ class CRM_CiviAwards_Service_AwardImport {
    *   Information for creating the award and related entities.
    */
   public function create(array $params) {
+    $this->validateParams($params);
+
     $tx = new CRM_Core_Transaction();
 
     $params['case_type_category'] = 'awards';
@@ -63,6 +65,18 @@ class CRM_CiviAwards_Service_AwardImport {
         ['name' => 'Application Manager'],
       ],
     ];
+  }
+
+  /**
+   * Validate information received.
+   *
+   * @param array $params
+   *   Information for creating the award and related entities.
+   */
+  private function validateParams(array $params) {
+    if (empty($params['title'])) {
+      throw new API_Exception('Invalid param received: Award Title should not be empty');
+    }
   }
 
 }
--- a/CRM/CiviAwards/Service/AwardImport.php
+++ b/CRM/CiviAwards/Service/AwardImport.php
@@ -17,7 +17,7 @@ class CRM_CiviAwards_Service_AwardImport {
     $tx = new CRM_Core_Transaction();
 
     $params['case_type_category'] = 'awards';
-    $params['definition'] = $this->createDefinition($params);
+    $params['definition'] = $this->createDefaultDefinition($params);
     try {
       $caseType = civicrm_api3('CaseType', 'create', $params);
     }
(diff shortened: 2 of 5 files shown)


api/v3/AwardImport.php
```diff
@@ -0,0 +1,105 @@
...
+
+use CRM_CiviAwards_Service_AwardImport as AwardImportService;
+
+/**
+ * AwardImport.create API specification.
+ *
+ * @param array $spec
+ *   Description of fields supported by this API call.
+ */
+function _civicrm_api3_award_import_create_spec(array &$spec) {
```
tunbola, line 16: I think what we need here for the fields is basically the Case Type fields and the Award Details Entity fields. We can leverage on Civi's API calls here

```php
  $caseTypeFields = civicrm_api3('CaseType', 'getfields', ['api_action' => 'get']);
  $awardDetailFields = civicrm_api3('AwardDetail', 'getfields', ['api_action' => 'get']);
  $spec = array_merge($caseTypeFields, $awardDetailFields)

```

ref: https://github.com/compucorp/uk.co.compucorp.civicase/blob/07b7746e2ad5af5fa6d5d3e74e3d74e9cc7f5c9f/api/v3/Case/Getdetails.php#L18

CRM/CiviAwards/Service/AwardImport.php
```diff
@@ -0,0 +1,128 @@
...
+  /**
+   * Create an Award and the related entities.
+   *
+   * @param array $params
+   *   Information for creating the award and related entities.
+   */
+  public function create(array $params) {
+    $tx = new CRM_Core_Transaction();
+
+    $params['case_type_category'] = $this->getCaseCategoryValueForAwards();
```
tunbola, line 19: We can simply do this: `$params['case_type_category'] = 'awards';` Civi will do the convert to the option value behind the scene. You can try this on the API explorer to see what I mean.

CRM/CiviAwards/Service/AwardImport.php
```diff
@@ -0,0 +1,128 @@
...
+    $params['definition'] = $this->createDefinition($params);
```
tunbola, line 20: Yep. We should keep this, this is a good one as the previous import was not importing the definitions successfully. I think we should just keep the `activity_types` and `caseRoles` defaults to create this column so that it can be populated from start and remove other inputs.

CRM/CiviAwards/Service/AwardImport.php
```diff
@@ -0,0 +1,128 @@
...
+    try {
+      $caseType = civicrm_api3('CaseType', 'create', $params);
+    }
+    catch (Exception $exception) {
+      $tx->rollback();
+
+      throw new API_Exception('Exception while saving the Case Type for Award: ' . $exception->getMessage());
+    }
+    $caseTypeId = array_shift($caseType['values'])['id'];
```
tunbola, line 29: You can simply do `$caseTypeId = $caseType['id'];`

CRM/CiviAwards/Service/AwardImport.php
```diff
@@ -0,0 +1,128 @@
...
+    try {
+      civicrm_api3('AwardDetail', 'create', $params);
+    }
+    catch (Exception $exception) {
+      $tx->rollback();
+
+      throw new API_Exception('Exception while saving the AwardDetail: ' . $exception->getMessage());
+    }
+
+    if (!empty($params['review_panel_title'])) {
```
tunbola, line 43: We can take out the review panel creation part like I mentioned in other comments. The idea is to be able to import an Award with basic details to get it working and other complex parts can be filled in later via the UI.

api/v3/AwardImport.php
```diff
@@ -0,0 +1,105 @@
...
+    'description' => CRM_CiviAwards_ExtensionUtil::ts('Whether the award detail is for a template or not'),
+  ];
+
+  $spec['award_manager'] = [
+    'type' => CRM_Utils_Type::T_STRING,
+    'title' => 'Award manager',
+    'description' => 'An array of Contact IDs',
+  ];
+
+  $spec['review_fields'] = [
```
tunbola, line 66: We should remove support for review fields too. It's complex to do via the CSV.

CRM/CiviAwards/Service/AwardImport.php
```diff
@@ -0,0 +1,128 @@
...
+    catch (Exception $exception) {
+      $tx->rollback();
+
+      throw new API_Exception('Exception while saving the Case Type for Award: ' . $exception->getMessage());
+    }
+    $caseTypeId = array_shift($caseType['values'])['id'];
+
+    $params['case_type_id'] = $caseTypeId;
+    $params['award_manager'] = !empty($params['award_manager']) ? json_decode($params['award_manager'], TRUE) : [];
+    $params['review_fields'] = !empty($params['review_fields']) ? json_decode($params['review_fields'], TRUE) : [];
```
tunbola, line 33: We should remove support for review fields

CRM/CiviAwards/Service/AwardImport.php
```diff
@@ -0,0 +1,128 @@
...
+    }
+    catch (Exception $exception) {
+      $tx->rollback();
+
+      throw new API_Exception('Exception while saving the Case Type for Award: ' . $exception->getMessage());
+    }
+    $caseTypeId = array_shift($caseType['values'])['id'];
+
+    $params['case_type_id'] = $caseTypeId;
+    $params['award_manager'] = !empty($params['award_manager']) ? json_decode($params['award_manager'], TRUE) : [];
```
tunbola, line 32: I think it will be more intuitive to allow a comma separated list of contact ids for the award_manager column. This is easier to add for a lay user than adding a json string in the csv.

If we have that. Then we can do

```php
$params['award_manager'] = !empty($params['award_manager']) ? explode(",", $params['award_manager']) : [];
```

tunbola (changes requested):

tunbola:
@lisandro-compucorp , 
- Can you add a GIF when you update the PR that shows how to import the Award and the import, and mapping phases on the screen?
- Can you add a bit more background to the overview, for example you mentioned that an error occurred when an Award Case Type was imported using the CSV importer, can you add specifics?
- Also you added a link to compucorp.atlassian.net/browse/CPS-411 , This repo is an open source repo and not everyone will have access to our Jira, so it's better to add the information that you want to be available in the PR description.

CRM/CiviAwards/Service/AwardImport.php
lisandro-compucorp: Ok! Done: https://github.com/compucorp/uk.co.compucorp.civiawards/pull/164/commits/35bc299fcf06151a2e88b22dd80cd5ea14683c24

CRM/CiviAwards/Service/AwardImport.php
lisandro-compucorp: Fixed: https://github.com/compucorp/uk.co.compucorp.civiawards/pull/164/commits/102906d9515218ea6d056dc187d940bd5e91df5a

CRM/CiviAwards/Service/AwardImport.php
lisandro-compucorp: I agree with that. Done here: https://github.com/compucorp/uk.co.compucorp.civiawards/pull/164/commits/98f1b35279127066c2ac190147e62247aabb22b4

api/v3/AwardImport.php
lisandro-compucorp: Removed: https://github.com/compucorp/uk.co.compucorp.civiawards/pull/164/commits/3fb7778d49adbfcc4221c455f095fac30cdd943c

api/v3/AwardImport.php
```diff
@@ -0,0 +1,105 @@
...
+    'description' => 'A two-dimensional array of activity types. Example [{"name": "Applicant Review"}, {"name": "Email"}]',
+  ];
+
+  $spec['statuses'] = [
+    'type' => CRM_Utils_Type::T_STRING,
+    'title' => 'Award statuses',
+    'description' => 'A two-dimensional array of Award statuses. Example ["won", "lost"]',
+  ];
+
+  $spec['review_panel_title'] = [
```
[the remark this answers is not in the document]
lisandro-compucorp: Removed https://github.com/compucorp/uk.co.compucorp.civiawards/pull/164/commits/d3355d367a9c4590ede329abc89c051ef00e893a

CRM/CiviAwards/Service/AwardImport.php
lisandro-compucorp: Ok, agree with that. I left only the default values here, and it the same result than creating the Award using the interface. Done here: https://github.com/compucorp/uk.co.compucorp.civiawards/pull/164/commits/e42e816414f9f5801468d3505ea7a1afc599db71

api/v3/AwardImport.php
lisandro-compucorp: I tried this, because sounds a good idea. I merged the values of the arrays, but I couldn't make it work finally.
The final array was similar to the one that I manually created, but with much more information on every row.
At the end, when I tried the process manually, it returned a 500 error, very difficult to debug (I guess because of the queue usage, was so difficult to see what was failing)
It was taking me too long, then I decided to move on with this. I removed the unused specs: https://github.com/compucorp/uk.co.compucorp.civiawards/pull/164/commits/fedf6f7677a992d1ef589724ea17e5b987d17849

lisandro-compucorp:
@tunbola Thanks for your comments Tunbola. I have applied the changes and also updated the PR description.
Let me know how is looking now.

CRM/CiviAwards/Service/AwardImport.php
```diff
@@ -0,0 +1,81 @@
...
+    return TRUE;
+  }
+
+  /**
+   * Get the case category value for Awards.
+   *
+   * @return int
+   *   The case category id for Awards.
+   */
+  public function getCaseCategoryValueForAwards() {
```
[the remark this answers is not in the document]
lisandro-compucorp: Moved to the test class: https://github.com/compucorp/uk.co.compucorp.civiawards/pull/164/commits/0d70b6a039d7b7798d301cad4db12ac6668e0834

CRM/CiviAwards/Service/AwardImport.php
```diff
@@ -0,0 +1,81 @@
...
+  /**
+   * Create a default definition for the award.
+   *
+   * Makes an array with all the minimal default information required for
+   * the definition field on the award.
+   *
+   * @return array
+   *   Array with details for the definition field.
+   */
+  private function createDefaultDefinition() {
```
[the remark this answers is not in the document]
lisandro-compucorp: Renamed: https://github.com/compucorp/uk.co.compucorp.civiawards/pull/164/commits/9e8f88f34e9f96a01fe182d9b0cc5c997181cc1d

tests/phpunit/CRM/CiviAwards/Service/AwardImportTest.php
```diff
@@ -0,0 +1,161 @@
...
+    $this->assertContains($contactIdTwo, $awardDetail['award_manager']);
+  }
+
+  /**
+   * Test that in case of error, all the operation is cancelled.
+   */
+  public function testErrorImportingAwardWithDetailInformation() {
+    $params = $this->getBaseParamsForAward();
+    $params['start_date'] = 'invalid date';
+    $this->expectException(API_Exception::class);
```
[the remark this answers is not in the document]
lisandro-compucorp: Done: https://github.com/compucorp/uk.co.compucorp.civiawards/pull/164/commits/175f1e4a98048385990c5e4f10d51b36ea9d1be3
(I was tempted to check with a regular expression the presence of some words that implies the error expected, but I think that this will be good enough)

api/v3/AwardImport.php
```diff
@@ -0,0 +1,74 @@
...
+    'title' => CRM_CiviAwards_ExtensionUtil::ts('Is Template?'),
+    'description' => CRM_CiviAwards_ExtensionUtil::ts('Whether the award detail is for a template or not'),
+  ];
+
+  $spec['award_manager'] = [
+    'type' => CRM_Utils_Type::T_STRING,
+    'title' => 'Award Manager',
+    'description' => 'A comma-separated list of Contact IDs',
+  ];
+}
```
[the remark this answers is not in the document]
lisandro-compucorp: Implemented here: https://github.com/compucorp/uk.co.compucorp.civiawards/pull/164/commits/109c30f27319c60f71d4f38c9578b0ca633141bf
This was failing because of, from the `getfields` API call, we got `award_subtype` field with an element named `pseudoconstant`. 
I don't know the exact reason but this was making the import to fail. Other fields also had the `pseudoconstant` key, but in them worked well, then is very specific of this only spec. 
I overwrite the value with the manual creation, without the problematic element.
Let me know if this sounds good.

api/v3/AwardImport.php
```diff
@@ -14,41 +14,9 @@
...
-  $spec['end_date'] = [
-    'type' => CRM_Utils_Type::T_DATE,
-    'title' => CRM_CiviAwards_ExtensionUtil::ts('End Date'),
-    'description' => CRM_CiviAwards_ExtensionUtil::ts('Award End Date'),
-  ];
+  $caseTypeFields = civicrm_api3('CaseType', 'getfields', ['api_action' => 'get']);
+  $awardDetailFields = civicrm_api3('AwardDetail', 'getfields', ['api_action' => 'get']);
+  $spec = array_merge($caseTypeFields['values'], $awardDetailFields['values']);
 
   $spec['is_template'] = [
```
tunbola, line 21: I think we can remove is_template? It should already be part of specs since it's part of Award detail fields?

tunbola (changes requested):

tunbola (approved):

CRM/CiviAwards/BAO/AwardDetail.php
```diff
@@ -81,6 +81,10 @@ private static function validateParams(array &$params) {
    *   Parameters.
    */
   private static function validateDates(array $params) {
+    if (empty($params['start_date'])) {
+      throw new Exception("Award Start Date should not be empty");
+    }
+
```
tunbola, line 88: I think we should add a new method `validateMandatory` that will cater for all parameters that are mandatory and needs to be validated. 

Something like this

```php
private static function validateMandatory($params) {
    $mandatoryFields = [
      'start_date' => 'Start Date'
      'award_subtype' => 'Award Subtype',
      'award_manager' => 'Award Manager'
    ];

    foreach($mandatoryFields as $field => $label) {
      if (empty($params[$field])) {
        throw new Exception( "The {$label} field should not be empty",);
      }
    }
  }


```

If we need to make any other parameter mandatory, we'll just add to the array

CRM/CiviAwards/BAO/AwardDetail.php
```diff
@@ -119,4 +120,28 @@ private static function validateReviewFields(array $params) {
...
+   *
+   * @param array $params
+   *   Request Parameters.
+   *
+   * @throws \Exception
+   */
+  private static function validateAwardManager(array $params) {
+    if (empty($params['award_manager'])) {
+      throw new Exception('Award Manager should not be empty');
+    }
```
tunbola, line 134: Let's move the code above to the proposed validatemandatory method. The logic below can be left since it validates valid award managers/contacts.

tunbola (changes requested):

CRM/CiviAwards/BAO/AwardDetail.php
lisandro-compucorp: Good idea 💪 . Done here: https://github.com/compucorp/uk.co.compucorp.civiawards/pull/164/commits/5e2bb6395c6dbf6c97f47799681da2ed3f00d192

tunbola (dismissed):

tunbola (approved):

james-compucorp:
Great contribution @lisandro-compucorp !

PR #165, opened 2020-12-30
RSE-1572: Fix Custom Fields Issue On Awards

shahrukh-compucorp:
## Overview
This pr fixes the issue with custom fields on an award configuration page. If there were more than 25 case types belonging to current case category instance then an error occurred on custom fields tab due to which users were not able to configure custom fields for an award.

## Before
There was an error message on the custom fields tab
image-20201207-165106

## After
The above mentioned issue has been resolved
Peek 2020-12-30 18-01

## Technical Details
There was an api call to fetch all the case types that belong to current case category instance in `CRM/CiviAwards/Helper/CaseTypeCategory.php`, the result of this api call was used to determine that if the current award is valid or not. This api call was missing limit parameter which by default is 25 so only first 25 records were being fetched. This pr adds the limit parameters to the api call with a value of 0 which means to fetch all records without any limit and introduction of this paramater fixes the above described issue.


lisandro-compucorp (approved):
Looks good to me 👍

PR #169, opened 2021-01-08
RSE-1034: Add filter to display disabled awards and applications

reneolivo:
## Overview
This PR adds a a "show disabled awards" filter. The filter will show awards, applications, and activities that would normally be hidden because they belong to a disabled award.

⚠️ Important: This PR implements the work done in https://github.com/compucorp/uk.co.compucorp.civiawards/pull/75 which had to be closed until a solution for disabled cases would be taken. 
This PR is also related to https://github.com/compucorp/uk.co.compucorp.civicase/pull/675

## Before
Screen Shot 2021-01-07 at 9 53 27 PM
the filter doesn't exist.

## After
pr
both applications shown here belong to disabled awards

## Technical Details

We use the `civicase-checkbox` directive introduced in https://github.com/compucorp/uk.co.compucorp.civicase/pull/675

Some of the changes introduced in https://github.com/compucorp/uk.co.compucorp.civiawards/pull/75  were not implemented in this PR because they were not necessary:

* Adding a relationship from the AwardManager entity to the CaseType entity. This was already done and querying awards by case type fields is possible.
* the `case_type_id.is_active` filter is only passed to the case filters object and not to the `AwardManager` endpoint because it would require refactoring `processMyAwardsFilter`. This is not necessary because this returns active and disabled awards belonging to a user and the case filter object will ensure that only disabled awards are returned.

### DIFF
--- a/ang/civiawards/dashboard/directives/more-filters-dashboard-action-button.controller.js
+++ b/ang/civiawards/dashboard/directives/more-filters-dashboard-action-button.controller.js
@@ -92,8 +92,14 @@
       processMyAwardsFilter(model.selectedFilters.awardFilter)
         .then(processAwardSubtypeFilters)
         .then(function (awardSubtypeIds) {
+          var isCaseTypeActive = model.selectedFilters.showDisabledAwards
+            ? '0'
+            : '1';
           var param = {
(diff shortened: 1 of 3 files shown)


deb1990 (commented):
Approved

PR #170, opened 2021-01-13
RSE-1459: Filter Unused Award Subtypes

shahrukh-compucorp:
## Overview
This pr filters out the unused award sub types while creating or updating a custom field set for award. As these sub types were not used yet so they resulted in empty case type for that custom field set due to which the custom field set was leaking into other case categories as well.

## Before
The above mentioned problem existed and as shown below all the available sub types were show in the custom field set form.
Screenshot from 2021-01-13 12-53-32

## After
The mentioned problem is resolved and only those sub types are show which has atleast one award attached to it.
Screenshot from 2021-01-13 12-52-38

## Technical Details
`getSubTypes` function in file `CRM/CiviAwards/Helper/CaseTypeCategory.php` was responsible for providing the award sub types, so it has been modified to provide only those award sub types which has been used atleast once.

### DIFF
--- a/CRM/CiviAwards/Helper/CaseTypeCategory.php
+++ b/CRM/CiviAwards/Helper/CaseTypeCategory.php
@@ -55,13 +55,27 @@ class CRM_CiviAwards_Helper_CaseTypeCategory {
   }
 
   /**
-   * Returns the option values for Award Subtypes.
+   * Returns the option values for Award Subtypes that has been used.
    *
    * @return array
    *   Award Subtypes.
    */
   public static function getSubTypes() {
-    return AwardDetail::buildOptions('award_subtype');
+    $usedSubTypes = civicrm_api3('AwardDetail', 'get', [
+      'return' => ['award_subtype'],
+      'options' => ['limit' => 0],
+    ]);
+    if (empty($usedSubTypes['values'])) {
+      return [];
+    }
+    $usedSubTypes = array_unique(array_column($usedSubTypes['values'], 'award_subtype'));
+
+    $allSubTypes = AwardDetail::buildOptions('award_subtype');
+    $allSubTypes = array_filter($allSubTypes, function ($key) use ($usedSubTypes) {
+      return in_array($key, $usedSubTypes);
+    }, ARRAY_FILTER_USE_KEY);
+
+    return $allSubTypes;
   }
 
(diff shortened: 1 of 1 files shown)


jamienovick:
Hey @shahrukh-compucorp @tunbola @james-compucorp not sure why we are doing this but it won't work as you will need to configure custom fields whether or not you have created a case type of that subtype. We need a different solution here.

tunbola:
@jamienovick , For adding a custom field to cases in Civi originally, one needs to select the case type or select the `Any` option. In this case when a subtype that has no case type is selected, there is no case type to add to the Db and we add NULL in the `extends_entity_column_value` column in the custom_group table which stores the case type IDs. This causes the custom group to leak to all cases in the system as NULL means it applies to all cases. 
The solution here was to prevent it leaking by making sure that a subtype without a case type is not shown. 
This seems to be the best approach and @shahrukh-compucorp  discussed this with PM before proceeding, not sure if you have another solution in mind?
cc @james-compucorp

CRM/CiviAwards/Helper/CaseTypeCategory.php
```diff
@@ -55,13 +55,27 @@ public static function getApplicantManagementCaseCategories() {
...
   public static function getSubTypes() {
-    return AwardDetail::buildOptions('award_subtype');
+    $usedSubTypes = civicrm_api3('AwardDetail', 'get', [
+      'return' => ['award_subtype'],
+      'options' => ['limit' => 0],
+    ]);
+    if (empty($usedSubTypes['values'])) {
+      return [];
+    }
+    $usedSubTypes = array_column($usedSubTypes['values'], 'award_subtype');
```
tunbola, line 71: This list could be pretty long and be full of duplicates for the award_subtypes values, we can reduce the duplicates at this point. 
```php
$usedSubTypes = array_unique(array_column($usedSubTypes['values'], 'award_subtype'));
```

tunbola (changes requested):

CRM/CiviAwards/Helper/CaseTypeCategory.php
shahrukh-compucorp: changed

tunbola (approved):

PR #172, opened 2021-01-15
RSE-1597: Block reviews for disabled applications

reneolivo:
## Overview
This PR blocks reviews from being added or updated for disabled applications.

This PR depends on https://github.com/compucorp/uk.co.compucorp.civicase/pull/679

## Before & After
Please look the last part of the civicase PR.

## Technical Details

* The "Add Review" and review action buttons are disabled when the case is disabled.
* Unit tests have been fixed as they were failing before. Civicase updated the way mocked case types are added here https://github.com/compucorp/uk.co.compucorp.civicase/pull/671/files#diff-6549489c078a5b7f418956858cebef0412da1eafa451159dd8e15c4e875db017R415-R421 because the way we were adding data was not consistent. We were passing arrays when we needed to be passing objects. To avoid having inconsistent behaviour only a single case type data object is accepted so we use `_.each` to pass all award case types.

### DIFF
--- a/ang/civiawards/reviews-tab/directives/reviews-case-tab-content.directive.html
+++ b/ang/civiawards/reviews-tab/directives/reviews-case-tab-content.directive.html
@@ -4,6 +4,7 @@
       class="btn btn-primary crm-popup civiawards__reviews-tab__action-button"
       ng-click="handleAddReviewActivity()"
       type="button"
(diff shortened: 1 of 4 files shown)


deb1990:
Also unit tests are failing, please check @reneolivo

deb1990 (approved):

PR #176, opened 2021-01-21
CPS-423: Sync Log Tables With Actual Tables While Running Upgraders

shahrukh-compucorp:
## Overview
This pr fixes the error which occurred during deployment due to the fact that some upgraders were modifying the tables but not applying the change to the log table.

## Before
Running this upgraders resulted in an error that halted the whole deployment process.

## After
The code for syncing the log tables with new changes have been added to the upgraders.

## Technical Details
`Step1007.php` and `Step1007.php` has been updated with the code for syncing the log tables.

### DIFF
--- a/CRM/CiviAwards/Upgrader/Steps/Step1007.php
+++ b/CRM/CiviAwards/Upgrader/Steps/Step1007.php
@@ -16,6 +16,7 @@ class CRM_CiviAwards_Upgrader_Steps_Step1007 {
   public function apply() {
     $this->renameDbColumn();
     $this->renameOptionGroup();
+    $this->syncLogTables();
 
     return TRUE;
   }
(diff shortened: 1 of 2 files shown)


CRM/CiviAwards/Upgrader/Steps/Step1007.php
```diff
@@ -52,4 +53,12 @@ private function renameDbColumn() {
     CRM_Core_DAO::executeQuery($renameColumnSql);
   }
 
+  /**
+   * Sync log tables.
+   */
+  private function syncLogTables() {
+    $logging = new CRM_Logging_Schema();
+    $logging->fixSchemaDifferences();
```
tunbola, line 61: I think `$logging->fixSchemaDifferencesFor('civicrm_civiawards_award_detail')` would be a better option since this will only fix schema differences for the Award detail table. `$logging->fixSchemaDifferences()` will attempt to fix schema differences for all tables even though it won't be created for those that already have but will be more optimised if we create for exact table.

tunbola (changes requested):

CRM/CiviAwards/Upgrader/Steps/Step1007.php
shahrukh-compucorp: yeah that is a good point, changed

shahrukh-compucorp: @tunbola can you review again please

tunbola (approved):

PR #178, opened 2021-01-26
RSE-1627: Fix award filter notifications

reneolivo:
## Overview
This PR fixes the notifications for the award filters on the dashboard. The notification red mark would stay on as soon as the filter's value changed and would not be removed, as per its normal behaviour, after reverting back the form to its original state.

It also updates the label for the disabled awards filter. The reason for the change is to make it explicit that only disabled awards and applications will be displayed instead of giving the impression of showing both enabled and disabled awards and applications.

This PR is related to https://github.com/compucorp/uk.co.compucorp.civicase/pull/687

## Before
pr

## After
pr

## Technical Details

* We updated the `isNotificationVisible` function in `more-filters-dashboard-action-button.controller.js ` so it includes the disabled award filter in the check.
* The filter was renamed to `onlyShowDisabledAwards`.
* The label for the filter was updated.


deb1990:
@reneolivo Code looks good, but can we fix the Overview section please. Right now its not clear how exactly notifications are fixed. So something like
"After XYZ PR, when applying Disabled Filters from More Filters popup, the red dot icon was not shown"

deb1990 (approved):

PR #179, opened 2021-01-28
ACU-441: Add Award Payments Form

tunbola:
## Overview
This PR adds the Award Finance Payments form

## Before
The Awards Finance Payment Form was not present

## After
The Awards Finance Payment Form was was added.
The Award Payment Form class was added with functionality for creating, viewing and editing an award payment. 

AwardPayments

## Notes
- There will be subsequent PR's to add the functionality for deleting an Award Payment and also to add logic around who can create, view or edit an Award payment and what fields can be edited.
- Basic unit tests will be added in subsequent PR's.
- There form still needs to be styled as some buttons are not properly styled and the form fields are not well aligned but will leave that to FE to style according to spec.

### DIFF
--- a/CRM/CiviAwards/Form/AwardPayment.php
+++ b/CRM/CiviAwards/Form/AwardPayment.php
@@ -1,0 +1,413 @@
+<?php
+
+use CRM_CiviAwards_ExtensionUtil as E;
+
+/**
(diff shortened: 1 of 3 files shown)


lisandro-compucorp (approved):
Looks good to me 👍

PR #182, opened 2021-02-04
ACU-422: Payments Tab

reneolivo:
## Overview
This PR adds the Payments tab that is only visible to case instances that support finances.

Only the base template is implemented and there are no functionalities associated to the tab yet.

As a note, the original specs mention that the tab should be "Finance" but this was updated since the only content will have for now are payments. This will change in the future, but for now we use payments.

## Before
This tab was not implemented.

### Case Instances that support finances
Screen Shot 2021-02-04 at 12 23 55 PM

### Case Instances that do not support finances

Screen Shot 2021-02-04 at 11 35 14 AM

## Technical Details

* A new `civiawards-payments-tab` angular module was created. This module contains all the changes required to display the payments tab.
* An `AddModule` hook was implemented to conditionally include the `civiawards-payments-tab` only for award instances that support finances.
* `add-payments-tab.config.js` adds the payments tab.
* `payments-case-tab.service.js` and `payments-case-tab-content.directive.js` add the basic template structure for the payments tab but they don't implement any logic yet.


jamienovick:
@reneolivo usual comments about shadows on panels and buttons. Can we get those automatically in the CSS in shroedtich?

reneolivo:
@jamienovick the task was not about adding all the details to the tab so the classes were not applied. Added them now so you could appreciate them better, but the proper implementation should be done when working on the task that is specific to the tables or the creating payments.

We have the classes defined in Civicase, not Shoreditch because this is very specific to civicase. We just need to add the classes to the proper elements.

deb1990 (changes requested):

deb1990 (approved):

CRM/CiviAwardsPaymentsTab/Hook/AddCiviCaseDependentAngularModules/AddModule.php
```diff
@@ -0,0 +1,53 @@
...
+
+use CRM_CiviAwards_Service_FinanceManagementSetting as FinanceManagementSettingService;
+
+/**
+ * Add Payments Tab module.
+ *
+ * This will add the payments tab module to awards instances that support
+ * finances.
+ */
+class CRM_CiviAwardsPaymentsTab_Hook_AddCiviCaseDependentAngularModules_AddModule {
```
tunbola, line 11: The class name seems a bit ambiguous, why don't we use rename to `PaymentsTabModule`. I removed the word `Add` because the folder name already describes that classes in this folder are adding to the civicase angular dependent modules.

tunbola (changes requested):

tunbola (approved):

PR #183, opened 2021-02-05
ACU-489: Award Payments Form Edit/View/Delete Business Logic

tunbola:
## Overview
This PR adds some business logic to the Awards Payments form. It adds logic that determines when a field is editable and who can edit it and also defines permissions for when an awards payment form can be deleted and who can delete. 

## Before
The business logic for editing/viewing/deleting an awards payment form was not present.

## After
**Edit Awards Payment FOrmBusiness Logic**
- If the status of the payment is in: "Exported, Paid, Failed" then the following fields are not editable by the user and are in a disabled state: Payee, Type, Payment amount, Payment currency, Due date, Payee ref. Note that the disabled fields are not validated on the backend, The purpose of the disabled fields just to denote the fact that they shouldn't be changed and they can easily be changed by simply changing the award payment status and re-submitting the form.

Edit Payment  CaseLatest 2021-02-08 10-36-22

**View Award Payments Form Business Logic**
- The edit button on the form is not shown if the Award payment status is Exported

View Payment  CaseLatest 2021-02-08 10-39-52

**Delete Award Payments Form Business Logic**
- The delete button is only shown if the Award Payments status is not in Exported, Paid or Failed and user has "delete activities" permission OR if the user has the "Administer CiviCRM permission"

## Comments
Unit tests for the Award Payments form was added. The Form calls some hooks already added in the extension and the tests was throwing some errors due to some non existent variables in the hooks. This was fixed in both this extension and the civicase extension since civicase needs to be installed for awards extension to run.

### DIFF
--- a/CRM/CiviAwards/Form/AwardPayment.php
+++ b/CRM/CiviAwards/Form/AwardPayment.php
@@ -477,11 +477,11 @@ class CRM_CiviAwards_Form_AwardPayment extends CRM_Core_Form {
       $this->getValueForActivityStatus('exported_complete'),
     ];
 
-    return in_array($activityStatus, $notEditableStatuses) ? TRUE : FALSE;
+    return in_array($activityStatus, $notEditableStatuses);
   }
 
   /**
-   * Get pre-defined fields that are to be un-editable.
+   * Get pre-defined fields that are un-editable.
    *
    * @return array
    *   Form fields.
--- a/CRM/CiviAwards/Form/AwardPayment.php
+++ b/CRM/CiviAwards/Form/AwardPayment.php
@@ -22,6 +22,13 @@ class CRM_CiviAwards_Form_AwardPayment extends CRM_Core_Form {
   private $activityId;
 
   /**
+   * Whether Activity Status is Exported or not.
+   *
+   * @var bool
+   */
+  private $isActivityStatusExported;
+
+  /**
    * Custom fields group tree.
    *
    * @var array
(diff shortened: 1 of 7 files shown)


CRM/CiviAwards/Form/AwardPayment.php
```diff
@@ -396,18 +416,107 @@ private function getAggregatedFieldsToInsertAfter() {
...
+   */
+  private function activityStatusIsLocked() {
+    $activityStatus = $this->activity['status_id'];
+    $notEditableStatuses = [
+      $this->getValueForActivityStatus('paid_complete'),
+      $this->getValueForActivityStatus('failed_incomplete'),
+      $this->getValueForActivityStatus('exported_complete'),
+    ];
+
+    return in_array($activityStatus, $notEditableStatuses) ? TRUE : FALSE;
```
lisandro-compucorp, line 480: No need for the ternary operator. Just return the result of the function.

CRM/CiviAwards/Form/AwardPayment.php
```diff
@@ -396,18 +416,107 @@ private function getAggregatedFieldsToInsertAfter() {
...
+  }
+
+  /**
+   * Get pre-defined fields that are to be un-editable.
```
lisandro-compucorp, line 484: Small typo here. Remove the "to be"

lisandro-compucorp (approved):
👍

PR #186, opened 2021-02-10
ACU-494: Add payment filters to the payment tab

reneolivo:
## Overview
This PR allows payment activities to be filtered.

This PR is related to https://github.com/compucorp/uk.co.compucorp.civicase/pull/699

## Before
Screen Shot 2021-02-09 at 9 41 14 AM
(filters don't exist)

## After
pr

## Technical Details

* The `civiawardsPaymentFilters` directive contains all the fields necessary for filtering and dropdown options, but does not do the actual filtering. 
* The payments filtering is done on the `civiawardsPaymentsTable` directive through the `filterPayments` function. On the directive's template we listen to the form submit event and trigger `filterPayments($filters)` where `$filters` are the values as provided by the `civiawardsPaymentFilters` directive.

# Filters style changes

## Before
Screen Shot 2021-02-16 at 4 41 00 PM

## After
Screen Shot 2021-02-16 at 4 38 21 PM

## Technical details

* Created a new sass file for the payments tab: `.civiawards__payments-tab`.
* The `!important` rules can't be avoided because they try to override inline styles added by scripts.
* The constant values can't be translated to variables because the don't use variables in their style definition. The `16px` belongs to a margin that the date inputs have between each other, the 2px is because of the input's borders, and the 8px is for the clear icon's height.


jamienovick:
@reneolivo can we do anything about the margin between the filter box and the payment table? I think it's all a bit on top of itself in the image so the shadow from the table seems to be over the filters box? I think the problem is that the shadow on the filters and on the table si kinda different so it looks out of place? Any thoughts on that... perhaps they need to be unified?

swastikpareek:
@reneolivo : Please add make the PR title a bit more descriptive `Add payment filters to the payment tab'
Also, please make sure the PR is aligned with the PR guidelines https://compucorp.atlassian.net/wiki/spaces/SD/pages/1520304185/Pull+Requests+guidelines

reneolivo:
@james-compucorp that shadow comes from the collapsible panel itself, not from the table's shadow overlapping. I moved the filters to a panel of its own so you can see it for yourself:
pr

In two different panels it looks very distracting with all those shadows flying around. Something I did fix was including the filters inside of the table's panel so the outer shadow covers the filters as well:

pr

I think this is the best outcome.

If we want to remove the shadows from the collapsible panel we need to do this through Shoreditch and that will remove it from everyone. I don't think we want to go that route.

What do you think @jamienovick ?

deb1990:
@reneolivo The Filter block does not have any "Filter" button, and filters are applied as you type. But in general in civicrm, filters always have a button. Some thing like this
2021-02-15 at 5 46 PM

deb1990:
Filter button looks good @reneolivo 
But the inputs still needs some fixing
2021-02-16 at 10 41 AM
This middle input is longer than the others, but all of them should have same width.

Also we almost forgot, lets add Backstop tests for this tab. Its fine if you want to add it in a future PR too :)

deb1990 (approved):

PR #189, opened 2021-02-15
ACU-502: Award Finance General Edits 

tunbola:
## Overview
This PR makes some adjustments to the Award payment form.
- The currency type field and the corresponding payment amount fields were aligned so that they are next to each other rather than being on separate lines in the form.
- The Value in Functional Currency Type custom field was renamed to Value in Functional Currency (Just the label) and the Value in Functional Currency field was renamed to Value in Functional Currency Amount (both label and name).
- The ability to configure/add to the currency type option values was also removed for currency type related fields.
- The time stamp for the due date field was also removed so the field is now date only.

Create New Payment  CaseLatest 2021-02-15 18-27-30

## Before
These changes were not present.

## After
- These changes are added.
Create New Payment  CaseLatest 2021-02-15 18-40-42

### DIFF
--- a/templates/CRM/CiviAwards/Form/AwardPayment.tpl
+++ b/templates/CRM/CiviAwards/Form/AwardPayment.tpl
@@ -5,10 +5,17 @@
             {if $elementName == 'separator'}
                 <hr>
             {else}
-              <div class="form-group row">
-                <label class="col-sm-2">{$form.$elementName.label}</label>
-                <div class="col-sm-10">{$form.$elementName.html}</div>
-              </div>
+                {if in_array($elementName, $fieldsAfterCurrencyTypes)}
+                    {$form.$elementName.html}</div></div>
+                {else}
+                  <div class="form-group row">
+                    <label class="col-sm-2">{$form.$elementName.label}</label>
+                    <div class="col-sm-10">{$form.$elementName.html}
+                    {if !in_array($elementName, $currencyTypeFields)}
+                    </div>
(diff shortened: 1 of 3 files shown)


jamienovick:
Thanks @tunbola 
A quick question - not sure if this got filtered through - is there any possibility of the field labels being dynamic and being picked up from the custom field labels (where they are displayed). This would allow us to tweak it for other clients? Just a question as the field labels are still confusing for some users I will think but the client required these labels.

templates/CRM/CiviAwards/Form/AwardPayment.tpl
```diff
@@ -5,10 +5,17 @@
             {if $elementName == 'separator'}
                 <hr>
             {else}
-              <div class="form-group row">
-                <label class="col-sm-2">{$form.$elementName.label}</label>
-                <div class="col-sm-10">{$form.$elementName.html}</div>
-              </div>
+                {if in_array($elementName, $fieldsAfterCurrencyTypes)}
```
lisandro-compucorp, line 8: Good idea 🙂

lisandro-compucorp (approved):
Looks good to me 👍

tunbola:
@jamienovick , The field label is dynamic as it is and always picked up from Custom field labels. We don't have any business logic based on the labels. The logic are based on the custom field names and this can't be changed from the UI. So the labels can be changed to whatever is suitable for the client.

PR #195, opened 2021-02-19
ACU-522: Add Euro currency

reneolivo:
## Overview
This PR adds the Euro to the list of currencies.

## Before
Screen Shot 2021-02-18 at 11 35 38 PM

## After
Screen Shot 2021-02-18 at 11 38 22 PM

## Technical Details

We added a new upgrader that uses `CRM_Core_BAO_OptionValue::ensureOptionValueExists` to add the Euro currency once no matter how many times the function is called.

### DIFF
--- a/CRM/CiviAwards/Upgrader/Steps/Step1010.php
+++ b/CRM/CiviAwards/Upgrader/Steps/Step1010.php
@@ -1,0 +1,28 @@
+<?php
+
+/**
+ * Adds the Euro currency option.
+ */
+class CRM_CiviAwards_Upgrader_Steps_Step1010 {
+
+  /**
+   * Apply function that runs on setup.
+   *
+   * @return bool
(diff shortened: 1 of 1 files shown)


jamienovick:
@reneolivo @james-compucorp I'm not sure we will want to do this. clients will each have their available currency configured for their entire CRM and we don't want to change these just because somebody enables awards finance to include €. We would simply want to to use the list of available currencies that they have configured for the CRM overall.

tunbola:
I don't think we need this ticket. I left a comment also on the ticket yesterday that the options are configurable and we don't need to do this. We should probably close this. cc @james-compucorp

james-compucorp:
@tunbola @reneolivo please leave this PR out as discussed. PM/ Client can configure specific currencies based on preference.

tunbola:
Not needed anymore

PR #201, opened 2021-02-23
ACU-532: Fix Award Manager Permission Issues With Award Payment

tunbola:
## Overview
When a user with the Award Manager role access the award payment listing page or the create/edit form popup, Previously  created award payments are not visible on the listing screen and also on the form, the custom fields belonging to the Award payment information are not visible. 

## Before
Payments Listing Screen
Manage Cases | CaseLatest 2021-02-24 13-00-20

Payment Create/Edit Screen
Manage Cases | CaseLatest 2021-02-24 13-00-05
 
## After
Payments Listing Screen
Manage Cases | CaseLatest 2021-02-24 13-03-10

Payment Create/Edit Screen
Manage Cases | CaseLatest 2021-02-24 13-03-28

## Technical Details
- The listing page is created from FE by using the CustomField API to fetch custom fields belonging to the Award Payments group. The permission required to access this API is the `administer CiviCRM and access all custom data` permissions. The Award manager would not have these permissions, so when the custom fields to be accessed is associated with the Payment Information activity type, the newly created `access payment custom field set` permission which the Award manager will have access to. Similar thing was done for Award review fields. 
-  For the Awards payment custom fields to show for the Award manager on the payment form, the Award manger needs to have an ACl that has access to these fields or have the ` access all custom data` permission, because the Award manager would not have access to these, a new hook `CRM_CiviAwards_Hook_AclGroup_AllowAccessToPaymentGroups` was added to allow the user with the `access payment custom field set` permission to have access to these groups as part of the allowed ACL groups for the user.

### DIFF
--- a/CRM/CiviAwards/Hook/AclGroup/AllowAccessToPaymentGroups.php
+++ b/CRM/CiviAwards/Hook/AclGroup/AllowAccessToPaymentGroups.php
@@ -1,0 +1,78 @@
+<?php
+
+use CRM_CiviAwards_Hook_AclGroup_ActivityTypeCustomGroupAccess as ActivityTypeCustomGroupAccess;
+use CRM_CiviAwards_Setup_CreateAwardPaymentActivityTypes as CreateAwardPaymentActivityTypes;
+use CRM_CiviAwards_Hook_AlterAPIPermissions_Award as AwardPermission;
+
+/**
+ * Class CRM_CiviAwards_Hook_AclGroup_AllowAccessToApplicantReviewGroups.
+ */
+class CRM_CiviAwards_Hook_AclGroup_AllowAccessToPaymentGroups extends ActivityTypeCustomGroupAccess {
+
+  /**
+   * Modifies ACL groups for user with access payment set permission.
+   *
+   * Allows a user with access payment set permission to have access to the
+   * payment information custom group and custom fields belonging to this group.
+   *
+   * @param string $type
+   *   The type of permission needed.
+   * @param int $contactID
+   *   The contactID for whom the check is made.
+   * @param string $tableName
+   *   The tableName which is being permissioned.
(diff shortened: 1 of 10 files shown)


CRM/CiviAwards/Hook/AclGroup/AllowAccessToPaymentGroups.php
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  /**
+   * Returns the payment activity type Id.
+   *
+   * @return int
+   *   Activity type Id.
+   */
+  private function getAwardPaymentActivityTypeId() {
+    try {
```
lisandro-compucorp, line 66: I would prefer to always return something:
```php
  $result = null;  
  try {
      $result = civicrm_api3('OptionValue', 'getsingle', [
        'option_group_id' => 'activity_type',
        'name' => CreateAwardPaymentActivityTypes::AWARD_PAYMENT_ACTIVITY_TYPE,
      ])['value'];
    }
    catch (Exception $e) {
}

return $result;
```
What do you think?

CRM/CiviAwards/Hook/AlterAPIPermissions/Award.php
```diff
@@ -195,4 +218,27 @@ private function getCustomGroupDetails($customGroup) {
...
+   * Returns the activity type Id.
+   *
+   * @param string $optionValueName
+   *   Option value name.
+   *
+   * @return int
+   *   Activity type Id.
+   */
+  private function getActivityTypeId($optionValueName) {
+    try {
```
lisandro-compucorp, line 231: Same here than in the other class. I think that can work anyway without returning null, but it seems better.

lisandro-compucorp (approved):
👍

PR #206, opened 2021-03-01
IO-143: Disable/Uninstall CiviAwards properly

ahed-compucorp:
## Overview
This PR fixes the issue when opening the "Custom Data" page after disabling the extension and hides the "Awards" menu item. The `500 Internal Server` error is because CiviCRM cannot start/bootstrap properly if one of the classes is missing, e.g. `require not_existed_file.php`. For our case, `CRM_CiviAwards_Helper_CaseTypeCategory` is not loadable when the extension is disabled.

### Before - when disabling the extension
Opening the "Custom Data" page `/civicrm/admin/custom/group?reset=1` *is not possible* while the extension is disabled.
Screenshot from 2021-03-01 16-21-22

### After - when disabling the extension 
Opening the "Custom Data" page `/civicrm/admin/custom/group?reset=1` *is possible* while the extension is disabled, also related custom groups - created by the user - are hidden.

Screenshot from 2021-03-05 10-16-34

### Before "Awards menu item" - when disabling the extension
Screenshot from 2021-03-05 10-08-42

### After "Awards menu item"  - when disabling the extension
Screenshot from 2021-03-05 10-09-32

## Technical Details
- Activated/Deactivated the custom field extend option group for the case
category according to the extension status. This solved the problem because CiviCRM only loads active `cg_extend_objects` [CustomGroup.php#L2028](https://github.com/civicrm/civicrm-core/blob/da176e2b330d0d7031c2840c3375c32a35683716/CRM/Core/BAO/CustomGroup.php#L2028).
- Activated/Deactivated "Award" menu item according to the extension status.


CRM/CiviAwards/Enable/ActivateCustomGroupSupportForAwardsCategory.php
```diff
@@ -0,0 +1,19 @@
...
+ */
+class CRM_CiviAwards_Enable_ActivateCustomGroupSupportForAwardsCategory {
+
+  /**
+   * Deletes the Awards option from the CG Extends option values.
+   */
+  public function apply() {
+    $caseCategoryCustomFieldExtends = new CaseCategoryCustomFieldExtends();
+    $caseCategoryCustomFieldExtends->activate(CaseTypeCategoryHelper::AWARDS_CASE_TYPE_CATEGORY_NAME);
+  }
```
tunbola, line 17: We need to approach this a bit differently. The issue is that the users can add more categories of type awards via the UI apart from those added via extension installation and each of these award categories will have corresponding option value entries in the `cg_extends` option group. 
We should probably do something like this:

```php
  public function apply() {
    $caseCategoryDetails = array_column(CRM_Civicase_Helper_CaseCategory::getCaseCategories(), 'name', 'value');
    $applicantCaseCategoryIds = CRM_CiviAwards_Helper_CaseTypeCategory::getApplicantManagementCaseCategories();
    $caseCategoryCustomFieldExtends = new CaseCategoryCustomFieldExtends();
    foreach ($applicantCaseCategoryIds as $caseCategoryValue) {
      $caseCategoryCustomFieldExtends->activate($caseCategoryDetails[$caseCategoryValue]);
    }
  }
```

we should do same for the deactivation too.

tunbola (changes requested):

CRM/CiviAwards/Enable/ActivateCustomGroupSupportForAwardsCategory.php
ahed-compucorp: Good catch!
Updated the PR to activated/deactivated any custom groups created by both the extension or the user.

tunbola: Looks like you misunderstood what I meant with this comment. We don't need any change in the civicase PR to accomplish what I said. Basically this class only takes care of the Awards case category which is installed in the extension by default and we need to do this too for any other case categories created by the user. Your initial code should work fine, we only need to do a loop to do same thing for the other case categories. the code I pasted earlier is still valid
```php
  public function apply() {
    $caseCategoryDetails = array_column(CRM_Civicase_Helper_CaseCategory::getCaseCategories(), 'name', 'value');
    $applicantCaseCategoryIds = CRM_CiviAwards_Helper_CaseTypeCategory::getApplicantManagementCaseCategories();
    $caseCategoryCustomFieldExtends = new CaseCategoryCustomFieldExtends();
    foreach ($applicantCaseCategoryIds as $caseCategoryValue) {
      $caseCategoryCustomFieldExtends->activate($caseCategoryDetails[$caseCategoryValue]);
    }
  }
```
If you can undo changes done in civicase and use this code, should work just fine.

CRM/CiviAwards/Enable/ActivateAwardsMenus.php
```diff
@@ -0,0 +1,19 @@
...
+ */
+class CRM_CiviAwards_Enable_ActivateAwardsMenus {
+
+  /**
+   * Activates the Awards menu items.
+   */
+  public function apply() {
+    $applicationMenu = new ApplicantManagementMenu();
+    $applicationMenu->activateItem(CaseTypeCategoryHelper::AWARDS_CASE_TYPE_CATEGORY_NAME);
+  }
```
tunbola, line 17: We need to disable not only award menus but also menus for case categories belonging to applicant management.
The following slightly modified code should do just fine. Same loops need to be in function to disable too.
```php

  public function apply() {
    $caseCategoryDetails = array_column(CRM_Civicase_Helper_CaseCategory::getCaseCategories(), 'name', 'value');
    $applicantCaseCategoryIds = CRM_CiviAwards_Helper_CaseTypeCategory::getApplicantManagementCaseCategories();
    $applicationMenu = new ApplicantManagementMenu();
    foreach ($applicantCaseCategoryIds as $caseCategoryValue) {
        $applicationMenu->activateItem($caseCategoryDetails[$caseCategoryValue]);
    }
  }
```

tunbola (changes requested):

CRM/CiviAwards/Disable/DeactivateCustomGroupSupportForApplicantManagement.php
```diff
@@ -0,0 +1,22 @@
...
+class CRM_CiviAwards_Disable_DeactivateCustomGroupSupportForApplicantManagement {
+
+  /**
+   * Deactivate the Award Sub Type option from the CG Extends option values.
+   */
+  public function apply() {
+    $caseCategoryCaseTypeCustomFieldExtends = new CaseCategoryCaseTypeCustomFieldExtends();
+    $caseCategoryCaseTypeCustomFieldExtends->deactivate(CaseTypeCategoryHelper::APPLICATION_MANAGEMENT_NAME);
+
+    CustomGroupsAPIHelpers::deactivateCustomGroupsForEntity(CaseTypeCategoryHelper::APPLICATION_MANAGEMENT_NAME . "Type");
```
tunbola, line 19: We don't need this line

tunbola (dismissed):

tunbola (dismissed):

omarabuhussein:
Update:

I rebased this with master, and added a new commit that removes the following custom groups when uninstalling the extension:

```
 'Payment_Approval', 'Awards_Payment_Bank_Details', 'Awards_Payment_Information'
```

So the extension does not complain when we try to install it again after uninstalling it that these custom groups are already exsit.

omarabuhussein:
passed testing, so I will merge.

PR #213, opened 2021-03-11
ACU-548: Fix Issue with Disappearing Review Fields

tunbola:
## Overview
Recently on a client site, intermittently, review fields would be unavailable when a reviewer tries to add a review. Clearing the cache would fix the issue but issue would re-appear again after some time.

## Before
Manage Applications  CaseLatest 2021-03-11 14-32-07

## After
The review fields appear and does not disappear intermitently
Manage Applications  CaseLatest 2021-03-11 14-33-57

## Technical Details
The function we are using to fetch profile fields for the Awards Review Form: `CRM_Core_BAO_UFGroup::getFields` sets profile custom fields in the cache [here](https://github.com/civicrm/civicrm-core/blob/master/CRM/Core/BAO/UFGroup.php#L729) so that next time the function fetches custom fields for a profile, the values are returned from the cache. 
The issue is that the function is used in a lot of places in core and the cache value can easily be overwritten because none of those places passes the `$ctype` variable which should have made the cache key unique. [For Example](https://github.com/civicrm/civicrm-core/blob/master/CRM/Event/Form/Registration.php#L516)

For example in this case, if an anonymous user visits an online event registration page, the profile custom fields is stored in cache with key `uf_group_custom_fields_` and because the anonymous user does not have access to some custom fields (awards review custom fields inclusive) these fields saved for the profile cache will be missing the awards review custom fields. The next time an administrator visits the award review form, these custom fields are fetched from the cache and the award custom fields are not returned as part of the results, hence the warning page that review fields are not available.

To fix this, the value of the `ctype` parameter: `AwardReview` is passed when requesting the profile custom fields. This will make the custom fields to be stored with a unique key and not be overwritten since the cache key is built with `$cacheKey = 'uf_group_custom_fields_' . $ctype;`

Setting the profile default values was also adjusted to use the fields fetched from this new cache key and setting default values was done using the `CRM_Core_BAO_UFGroup::setComponentDefaults` function.

### DIFF
--- a/CRM/CiviAwards/Form/AwardReview.php
+++ b/CRM/CiviAwards/Form/AwardReview.php
@@ -30,6 +30,13 @@ class CRM_CiviAwards_Form_AwardReview extends CRM_Core_Form {
   private $caseId;
 
   /**
+   * Store form fields.
+   *
+   * @var array
+   */
+  private $fields;
+
+  /**
    * Activity Id.
    *
    * @var int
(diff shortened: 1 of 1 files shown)


lisandro-compucorp (approved):
Looks good to me 👍

PR #220, opened 2021-04-15
RSE-1276: Update menu items setup to use singular labels

reneolivo:
## Overview
This PR updates the menu setup code because of changes done in https://github.com/compucorp/uk.co.compucorp.civicase/pull/760 

## Before & After
This does not change anything visually, only the code is updated so it follows the changes done in https://github.com/compucorp/uk.co.compucorp.civicase/pull/760

## Technical Details

`ApplicantManagementMenu`  class  extends `CRM_Civicase_Service_CaseCategoryMenu` and this class was updated to accept the name, label, and singular label to use when creating case category menu items. We now pass the missing arguments.

### DIFF
--- a/CRM/CiviAwards/Helper/CaseTypeCategory.php
+++ b/CRM/CiviAwards/Helper/CaseTypeCategory.php
@@ -13,6 +13,16 @@ class CRM_CiviAwards_Helper_CaseTypeCategory {
   const AWARDS_CASE_TYPE_CATEGORY_NAME = 'awards';
 
   /**
+   * Award category label.
+   */
+  const AWARDS_CASE_TYPE_CATEGORY_LABEL = 'Awards';
+
+  /**
+   * Award category singular label.
+   */
+  const AWARDS_CASE_TYPE_CATEGORY_SINGULAR_LABEL = 'Award';
+
+  /**
    * Applicant management instance name.
    */
   const APPLICATION_MANAGEMENT_NAME = 'applicant_management';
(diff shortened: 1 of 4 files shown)


deb1990 (dismissed):

deb1990:
@reneolivo Code looks okay, but please fix unit tests.

CRM/CiviAwards/Setup/ProcessAwardsCategoryForCustomGroupSupport.php
```diff
@@ -22,7 +22,10 @@ public function apply() {
       self::AWARDS_CATEGORY_CG_LABEL,
       'CRM_CiviAwards_Helper_CaseTypeCategory::getSubTypes;'
     );
-    $caseCategoryCustomData->create(CaseTypeCategoryHelper::AWARDS_CASE_TYPE_CATEGORY_NAME);
+    $caseCategoryCustomData->create([
+      'name' => CaseTypeCategoryHelper::AWARDS_CASE_TYPE_CATEGORY_NAME,
+      'label' => CaseTypeCategoryHelper::AWARDS_CASE_TYPE_CATEGORY_NAME,
+    ]);
   }
```
tunbola, line 29: Looks like this method only accepts a single parameter of the case category name and not an array. Can you double check?

tunbola (changes requested):

CRM/CiviAwards/Setup/ProcessAwardsCategoryForCustomGroupSupport.php
reneolivo: @tunbola thanks a lot.

I noticed that when creating the menu we use the case category name instead of the label. Probably the reason why we limit the label used. Wanted to have a middle ground in order to fix that issue in the future, but realised all the other extensions depended on this and went back. Should have updated this part.

tunbola (dismissed):

tunbola (approved):

PR #221, opened 2021-04-16
CPS-533: Show all review fields

deb1990:
## Overview
While creating awards, the "Add Review Field" popup, only used to show 25 custom fields, now it shows all of them.

## Before
Only 25 records were shown.

## After
All records were shown.

## Technical Details
Added `options: { limit: 0 }` to the api call in `review-fields-table.directive.js`.
Also replaced all instances of `crmApi` with `civicaseCrmApi`.

### DIFF
--- a/ang/civiawards/award-creation/directives/award.directive.js
+++ b/ang/civiawards/award-creation/directives/award.directive.js
@@ -16,8 +16,8 @@
   });
 
   module.controller('CiviAwardCreateEditAwardController', function (
-    $location, $q, $scope, $window, CaseTypeCategory, CaseStatus, crmApi, crmStatus,
-    Select2Utils, ts) {
+    $location, $q, $scope, $window, CaseTypeCategory, CaseStatus,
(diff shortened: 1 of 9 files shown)


lisandro-compucorp (approved):
👍

jamienovick:
@deb1990 @tunbola another example of the '25' limit...

deb1990:
> @deb1990 @tunbola another example of the '25' limit...

@jamienovick Right. Previously the approach we discussed, had some flaws so it was not possible to apply this. But @tunbola will bring this up in the Tech leads meeting, to get ideas.

tunbola:
@deb1990 , I discussed this during the meeting and we agreed to go ahead with the solution originally agreed in the email thread. So we should coordinate with James to add a ticket for the work and add it on the board. 

cc @jamienovick

lisandro-compucorp (approved):

lisandro-compucorp (approved):

PR #222, opened 2021-04-20
CPS-556: Fix unit tests

reneolivo:
## Overview
This PR fixes Javascript unit tests after they broke from changes done in the Civicase repo.

## Before
Screen Shot 2021-04-20 at 12 27 29 PM

## After
Screen Shot 2021-04-20 at 12 26 44 PM

## Technical Details

The review panels directive spec file was failing due to two changes done to the relationship types mock data files in civicase:

* https://github.com/compucorp/uk.co.compucorp.civicase/commit/daf2749c0a32987f85746a803f4bce06c894ef00#diff-90f2ea50f82e507ff64d5ac2572bc9aff3d6e6985e99e1690cc084b6f3065401R68-R86 this changed the mock data from a straight object, to a service that returns copies of the relationship types that can be used in unit tests. We changed `RelationshipTypeData.values` to `RelationshipTypeData.getSequential()`. `getSequential` already returns a copy of the objects so there was no need to clone them.
* https://github.com/compucorp/uk.co.compucorp.civicase/commit/cecd54dff8ac4e24bc4012cef3f0fda9226b5a5e#diff-90f2ea50f82e507ff64d5ac2572bc9aff3d6e6985e99e1690cc084b6f3065401R25-R48 added a couple of relationship types since the new type IDs were lower than the existing ones we were getting them first. To fix this issue we updated the expected relationship types. We don't need to check for specific ones so this change is ok.

### DIFF
--- a/ang/test/civiawards/award-creation/directives/review-panels/review-panels.directive.spec.js
+++ b/ang/test/civiawards/award-creation/directives/review-panels/review-panels.directive.spec.js
@@ -723,7 +723,7 @@
         is_error: 0,
         version: 3,
         count: 8,
-        values: _.cloneDeep(_.slice(RelationshipTypeData.values, 0, 4))
+        values: _.slice(RelationshipTypeData.getSequential(), 0, 4)
       };
     }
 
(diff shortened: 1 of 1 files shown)


deb1990:
This is already fixed in https://github.com/compucorp/uk.co.compucorp.civiawards/pull/221. Hence closing this. cc @reneolivo

deb1990 (approved):

PR #223, opened 2021-04-29
CPS-559: Handle Missing Data in Review Panel

lisandro-compucorp:
## Overview
This PR solves the problem observed on Award's Review Panels, when one or more of the related contacts groups, tags, or statuses were deleted.
In these cases, the review panel was not visible on the Award review screen.

## Before
When the review panel was referencing a deleted contact group, this error was displayed on the frontend code:
image
preventing the panel rendering. 

## After
The panel is correctly displayed.
image

## Technical Details
- A method for filtering the invalid or obsolete data was added. The method is called when first loaded the Panel, then data is cleaned on the presentation, and also if an edit action occurs, it is saved without this information.
- For doing the unit test, it was needed to make available the method `formatReviewPanelDataForUI`, since is the one in charge of sanitizing the panel, and is called before the edit activity.

### DIFF
--- a/ang/civiawards/award-creation/directives/review-panels/review-panels.directive.js
+++ b/ang/civiawards/award-creation/directives/review-panels/review-panels.directive.js
@@ -278,11 +278,38 @@
       var reviewPanelDataCopied = _.clone(reviewPanelData);
 
       _.each(reviewPanelDataCopied, function (reviewPanel) {
+        reviewPanel = removeObsoleteData(reviewPanel);
+
         reviewPanel.formattedContactSettings = getFormattedContactSettings(reviewPanel);
         reviewPanel.formattedVisibilitySettings = getFormattedVisibilitySettings(reviewPanel);
       });
 
       return reviewPanelDataCopied;
+    }
+
+    /**
+     * Remove obsolete data from the review panel.
+     *
+     * @param {Array} reviewPanel list of review panels fetched from API
+     * @returns {Array} list of review panels
+     */
+    function removeObsoleteData (reviewPanel) {
+      reviewPanel.visibility_settings.application_status = _.filter(reviewPanel.visibility_settings.application_status, function (statusId) {
+        return !!caseStatusesIndexed[statusId];
+      });
+      reviewPanel.visibility_settings.application_tags = _.filter(reviewPanel.visibility_settings.application_tags, function (tagId) {
+        return !!tagsIndexed[tagId];
+      });
+
+      reviewPanel.contact_settings.include_groups = _.filter(reviewPanel.contact_settings.include_groups, function (groupId) {
+        return !!groupsIndexed[groupId];
+      });
+
+      reviewPanel.contact_settings.exclude_groups = _.filter(reviewPanel.contact_settings.exclude_groups, function (groupId) {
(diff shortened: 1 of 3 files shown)


ang/civiawards/award-creation/directives/review-panels/review-panels.directive.js
```diff
@@ -556,8 +560,12 @@
         is_active: $scope.currentReviewPanel.isEnabled ? '1' : '0',
         case_type_id: $scope.awardId,
         contact_settings: {
-          exclude_groups: $scope.currentReviewPanel.contactSettings.groups.exclude,
-          include_groups: $scope.currentReviewPanel.contactSettings.groups.include,
+          exclude_groups: $scope.currentReviewPanel.contactSettings.groups.exclude.filter(function (item) {
```
deb1990, line 563: Code looks good @lisandro-compucorp , but we need to add unit tests to cover all these scenarios.

deb1990 (changes requested):

ang/civiawards/award-creation/directives/review-panels/review-panels.directive.js
lisandro-compucorp: @deb1990 something like this would be ok? https://github.com/compucorp/uk.co.compucorp.civiawards/pull/223/commits/0ebbd4dca0a50a3cd3e71fb3692649ed40bc34f4

ang/test/civiawards/award-creation/directives/review-panels/review-panels.directive.spec.js
```diff
@@ -209,6 +209,61 @@
         });
       });
 
+      describe('when saving a review panel with missing groups', () => {
```
deb1990, line 212: The code was added inside `getFormattedVisibilitySettings` function. Which is called during loading the data, and not saving. So we need to test the `getFormattedVisibilitySettings` function and not `saveButtonClickHandler`.

So it should be something similar to `when editing an existing review panel` test suite.

deb1990 (changes requested):

ang/test/civiawards/award-creation/directives/review-panels/review-panels.directive.spec.js
lisandro-compucorp: The method that filter the deleted groups (which is the new behaviour, the other is just to check the element of the array) is `getParamsForSavingReviewPanel`, that's why I tested the `saveButtonClickHandler`.

ang/civiawards/award-creation/directives/review-panels/review-panels.directive.js
```diff
@@ -41,6 +41,7 @@
     $scope.handleActiveStateReviewPanel = handleActiveStateReviewPanel;
     $scope.getActiveStateLabel = getActiveStateLabel;
     $scope.selectTab = selectTab;
+    $scope.formatReviewPanelDataForUI = formatReviewPanelDataForUI;
```
deb1990, line 44: This is not used anywhere. Do we need it?

deb1990 (changes requested):

deb1990 (approved):

PR #226, opened 2021-05-07
CPS-597: Fix Issue With Payment Tab Not Enabled When Award finance Is Activated.

tunbola:
## Overview
When the `Enable Finance Management` setting is turned on for an Applicant management case category, the payment tab on the applications for that award is not displayed. 

## Before
The issue described above exists:
Manage Applications  CiviQA 2021-05-07 16-58-12

## After
When the `Enable Finance Management` setting is turned on for an Applicant management case category, the payment tab on the applications for that award is now displayed.

Manage Applications  CiviQA 2021-05-07 16-54-53

## Technical Details
This issues is a regression caused by recent changes due to changing the case category name in the URL to the category value. 
The `AddCiviCaseDependentAngularModules_PaymentsTabModule` hook class responsible for adding the payment module for an award is not ran because the `shouldRun` method is relying on getting the case category name from the URL. The fix was to refactor the function to use the category value.


lisandro-compucorp (approved):
👍

PR #227, opened 2021-05-12
RSESPRT-43: Fix case types being overriddden by award

deb1990:
## Overview
Case type configurations was getting wiped when the Award Type is saved. This PR fixes that.

## Before
Case Type Configurations were overridden when Award was Saved.

## After
Case Type Configurations are not overridden when Award is Saved.

## Technical Details
In `ang/civiawards/award-creation/directives/award.directive.js`, we now keep a backup of the `CaseType` when editing, and assign it back when saving.


lisandro-compucorp (approved):
👍

PR #232, opened 2021-06-28
RSESPRT-61: Show all review fields

deb1990:
## Overview
The Review Fields table only shows 25 records, this PR fixes the same.

## Before
Only 25 records were shown.

## After
All records are shown.

## Technical Details
In `CRM/CiviAwards/Service/AwardProfile.php`, added `'options' => ['limit' => 0],`, so that `AwardDetail` returns all review fields.


tunbola (approved):

jamienovick:
Another 25 API limit? @deb1990 do we not have a global fix for this yet? @omarabuhussein @erawat @ana-compucorp fyi

deb1990:
@jamienovick Unfortunately no. We have discussed the same in our standup today itself, but it was confirmed by James that we dont have capacity to work on the global fix at this moment.

PR #250, opened 2022-07-19
RSE-1763: Allow editing review when status is draft

erawat:
## Overview

This PR adds ability to save draft 

## Before

- Save Draft and Save & Continue buttons did not exist. 
- Update action was not allowed on SSP screen. 
- User can only add a new review on SSP screen.

screenshot-compuclient-rse-1743-k8e cc-test site-2022 07 19-13_51_21

## After

- Save Draft and Save & Continue buttons have been added.
- Update action is allowed when the activity is draft. 

rse-1673

## Technical Details

- There is an issue with the margin between Cancel and Save Draft buttons, and this issue will need to be addressed in a separate PR in the _ssp_bootstrap_ theme.
- The status field only requires to display on the Admin screen, since SSP screen and admin screen are sharing the same form, the status is only added to the form when is not loading SSP and application will determine activity status based on the submitted buttons if the submission is from SSP screen.

### DIFF
--- a/CRM/CiviAwards/Form/AwardReview.php
+++ b/CRM/CiviAwards/Form/AwardReview.php
@@ -93,7 +93,7 @@ class CRM_CiviAwards_Form_AwardReview extends CRM_Core_Form {
     $hasDefaultValues = !empty($this->defaultValues);
 
     if ($hasDefaultValues) {
-      return;
+      return NULL;
     }
 
     $this->defaultValues = [];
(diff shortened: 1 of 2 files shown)


CRM/CiviAwards/Form/AwardReview.php
```diff
@@ -251,21 +242,95 @@ public function postProcess() {
...
+   *
+   * @return string
+   *   Status text ether Submitted or Updated.
+   */
+  private function getSessionStatusText() {
+    $submittedStatus = 'Submitted';
+    if ($this->isReviewFromSsp()) {
+      return $submittedStatus;
+    }
+    else {
```
olayiwola-compucorp, line 286: not important, but we can remove the else block, and bring the return out

olayiwola-compucorp (approved):

olayiwola-compucorp (approved):

PR #275, opened 2023-10-03
CIWEMB-435: Allow using multiple review custom groups

omarabuhussein:
## Before

When adding "review" fields to an Award type, the fields that appear are coming from a custom group called "Applicant Review Scoring Fields" that is shipped by default with CiviAwards:

2023-10-05 17_15_39-Custom Data _ Site-Install

here I've added 3 fields to this custom group:

image

and here you can see the above fields appear when trying to add review fields to an award type:

image

Civiawards is hardcoded to only use the above custom group, but sometimes we have clients who have large number of review fields (e.g 200 review) that they use for various workflows and processes, and given each custom field represnts a database column in the parent custom group table, the table size will exceed the limit of MySQL which brings down the site and result in various problems.

## After

Here I am adding the ability to configure new custom groups where the fields of such groups will appear when trying to add review fields, for such custom groups to be qualified they have to be used only for "Applicant Review" activity type. The benefit of this is now clients can split their review fields into separate smaller groups, so we don't end up with single big custom group. 

Here I've added two extra custom groups were both are used for Activities of "Applicant Review" type:

2023-10-05 17_29_05-Custom Data _ Site-Install

where the first extra group contains the following fields:

image

and the other contains this fields:

image

Here in the award type configuration page we can see these new fields:

2023-10-05 17_30_51-Configure Award - Gold _ Site-Install

and once added to the award type, they will appear when trying to add a review to any award of such type:

image

image

image

## Technical Details

I mostly altered the existing API calls that are hardcoded to fetch custom fields that belong to the   "Applicant Review Scoring Fields" (the machine name of this group is `Applicant_Review`) to instead fetch custom fields that belong to any custom group that extends activities of type "Applicant Review". I fetch the list of such custom groups using API v4 which unlike API v3 allows me to do a join on the option values table which allows me to able to detect such groups. Such API calls exist  in two places, the first one is inside the file `ang/civiawards/award-creation/directives/review-fields/review-fields-table.directive.js` which is responsible for the "Review" tab inside the Award type configuration page, and the other is `ang/civiawards/reviews-tab/directives/reviews-case-tab-content.directive.js`   which is responsible for showing the values in the award view page and the "add review" form.

## Update 1

As part of the discussion on the ticket, we decided that it would be more clear if we include the custom group name for each review field when configuring an award type, so I've added this commit to achieve that: https://github.com/compucorp/uk.co.compucorp.civiawards/pull/275/commits/35bf383222f8d741924c78fa5856467091855bcc

Here is how now the "Review fields" tab look like inside the awards type configuration page:

2023-11-07 20_27_54-Configure Award - Gold _ Site-Install

and here after clicking on "Add review field":

2023-11-07 20_28_56-Configure Award - Gold _ Site-Install

### DIFF
--- a/ang/civiawards-base.ang.php
+++ b/ang/civiawards-base.ang.php
@@ -46,5 +46,6 @@ return [
   'settings' => $options,
   'requires' => [
     'civicase-base',
+    'api4',
   ],
 ];
--- a/ang/civiawards.ang.php
+++ b/ang/civiawards.ang.php
@@ -29,6 +29,7 @@ return [
     0 => 'css/civiawards.min.css',
   ],
   'requires' => [
+    'api4',
     'crmUi',
     'crmUtil',
     'ngRoute',
--- a/ang/civiawards/award-creation/directives/review-fields/review-field-selection.html
+++ b/ang/civiawards/award-creation/directives/review-fields/review-field-selection.html
@@ -18,6 +18,7 @@
     <thead>
     <tr>
       <th>Review Field Name</th>
+      <th>Field Set</th>
       <th>Enabled</th>
       <th>Data Type</th>
       <th>Field Type</th>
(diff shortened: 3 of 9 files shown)


erawat (approved):
@omarabuhussein  nice approach/solution!

PR #278, opened 2024-05-13
RSESPRT-246: Prevent user from submitting review multiple times

olayiwola-compucorp:
## Overview
This PR disables the submit button post-click so that the user is prevented from multiple click/submission as this ends up breaking the form submission.

Also, handle error when a user attempts to submit a review without an activity ID.

## Before
**User can click on award submit multiple times**

## After
**User can only click on award submit once**
screen_recording_after

## Before
Drupal white page error is displayed if the user attempts to submit a review without an activity ID
Screenshot 2024-05-31 at 12 08 37

## After
Custom error message is displayed if the user attempts to submit a review without an activity ID

### DIFF
--- a/CRM/CiviAwards/Form/AwardReview.php
+++ b/CRM/CiviAwards/Form/AwardReview.php
@@ -117,36 +117,53 @@ class CRM_CiviAwards_Form_AwardReview extends CRM_Core_Form {
    * {@inheritDoc}
    */
   public function preProcess() {
-    if ($this->_action & CRM_Core_Action::ADD) {
-      $this->caseId = CRM_Utils_Request::retrieve('case_id', 'Positive', $this);
-    }
-    else {
-      $this->activityId = CRM_Utils_Request::retrieve('id', 'Positive', $this);
-      $this->activity = $this->getActivity();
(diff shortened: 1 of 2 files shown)


shahrukh-compuco (dismissed):

jamienovick:
@olayiwola-compucorp the after image for the first part of this ticket is a little unclear. Is it just the save button that is disabled or the whole screen as the whole screen appears to have an overlay?

shahrukh-compuco:
> @olayiwola-compucorp the after image for the first part of this ticket is a little unclear. Is it just the save button that is disabled or the whole screen as the whole screen appears to have an overlay?

yes its a full screen overlay

jamienovick:
Would it not be preferable to just make the buttons on the screen disabled after one of them is clicked? The overlay feels like it would be somewhat jarring from a UX perspective? @olayiwola-compucorp

shahrukh-compuco:
> Would it not be preferable to just make the buttons on the screen disabled after one of them is clicked? The overlay sounds like it would be somewhat jarring from a UX perspective? @olayiwola-compucorp

@jamienovick I have change it to just the form being disabled

shahrukh-compuco (approved):

PR #289, opened 2025-01-26
RSESPRT-319: Allow Access To Review Custom Groups For Users With Review permission

shahrukh-compuco:
## Overview
Allow users with award review permission to access custom groups that are configured to be used as review fields.

## Before
screen_recording_before

## After
screen_recording_after

### DIFF
--- a/civiawards.php
+++ b/civiawards.php
@@ -23,7 +23,19 @@ function civiawards_civicrm_config(&$config) {
 
   Civi::dispatcher()->addListener(
     'civi.api.authorize',
-    ['CRM_CiviAwards_Event_Listener_AlterPermission', 'authorize'],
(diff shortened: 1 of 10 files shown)


civiawards.php
```diff
@@ -23,7 +23,19 @@ function civiawards_civicrm_config(&$config) {
 
   Civi::dispatcher()->addListener(
     'civi.api.authorize',
-    ['CRM_CiviAwards_Event_Listener_AlterPermission', 'authorize'],
```
olayiwola-compucorp, line 26: Can we mention why we no longer use this `CRM_CiviAwards_Hook_AlterAPIPermissions_Award` and can we delete it, if it's no longer needed. @shahrukh-compuco

shahrukh-compuco: @olayiwola-compucorp I have just renamed it to reflect what it does

olayiwola-compucorp (approved):

PR #295, opened 2025-07-03
CIVIMM-340: Resolve Custom Group Access Issues for Users with Review Permissions

olayiwola-compucorp:
## Overview
Users with the `access review custom field set` permission were unable to access other non-application review custom group and fields on contact records, even when they had legitimate access through other means (ACL groups, role-based permissions, etc.). This occurred because our permission hook was overly restrictive and interfered with CiviCRM's native permission system.

### Root Cause
The issue was introduced in [PR #289](https://github.com/compucorp/uk.co.compucorp.civiawards/pull/289), where we added a global hook that intercepted **all** APIv4 custom group requests. The hook only granted access to users with either:
- Admin access + `access all custom data` permission, OR  
- `access review custom field set` permission for review-specific requests

This approach incorrectly **blocked** users who had legitimate access through other CiviCRM permission mechanisms (ACLs, roles, etc.).

### Solution
Instead of using a global hook that interferes with all custom group API calls, we've created a dedicated API wrapper specifically for applicant review custom groups.

#### Key Changes:
1. **New API Entity**: `\Civi\Api4\ApplicantReviewField::get()`
   - Provides direct access to applicant review custom groups
   - Eliminates the need for global hooks on standard custom group APIs
   - Respects CiviCRM's native permission system

2. **Simplified Permission Logic**:
   - Checks if user has `access review custom field set` permission
   - If yes: grants access to review custom groups
   - If no: delegates to CiviCRM's standard permission system
   - No longer interferes with other custom group access patterns

3. **Updated Frontend**:
   - Angular directive now uses the new `ApplicantReviewField` API
   - Maintains same functionality with improved permission handling

### Benefits
- ✅ **Fixes access issues**: Users with ACL/role permissions can now access custom fields normally
- ✅ **Maintains security**: Review custom groups still require appropriate permissions
- ✅ **Reduces complexity**: Eliminates global hooks and their potential side effects
- ✅ **Better performance**: No longer processes every custom group API call
- ✅ **Follows CiviCRM best practices**: Uses dedicated API entities instead of global hooks

### Testing
- [x] Users with `access review custom field set` can access review custom groups
- [x] Users with ACL permissions can access their permitted custom groups
- [x] Users without permissions are properly denied access
- [x] No interference with standard custom group/field operations


erawat (approved):

marcelocompucorp:
hi @shahrukh-compuco , as discussed, can you please check this PR, which was approved but not merged? cc @erawat-compuco

